PR #6, opened 2026-07-02
Implement WhatsAppCloudAdapter against the official Meta Cloud API

### LINKED ISSUE #4: Proposal: implement WhatsAppCloudAdapter (official Meta Cloud API)
## Problem

The WhatsApp side of the bot currently only runs on `BaileysAdapter`
(`src/platforms/whatsapp/baileysAdapter.ts`), which uses the unofficial
linked-device protocol. README and ARCHITECTURE.md both flag this explicitly:
"Baileys ... violates WhatsApp ToS — the number can be banned" and list the
official Cloud API as "the documented upgrade path." `WhatsAppCloudAdapter`
(`src/platforms/whatsapp/cloudAdapter.ts`) exists only as a 52-line stub where
every method throws "not implemented yet."

For a community-facing bot expected to run continuously, a single ToS
violation report can take out the WhatsApp channel entirely with no warning
and no clean recovery path (numbers are hard to get unbanned). This is the
single largest operational risk called out in the project's own docs, and the
seam to fix it (`PlatformAdapter`) is already designed and unused.

## Proposed approach

Implement `WhatsAppCloudAdapter` against the official WhatsApp Business
Platform Cloud API, hosted by Meta (no third-party BSP needed to start):

1. **Webhook intake** — a small HTTPS endpoint (reuse the existing Node
   service; add an Express/native `http` route) implementing:
   - `GET` verification handshake (`hub.mode`, `hub.verify_token`,
     `hub.challenge`) against `WHATSAPP_CLOUD_VERIFY_TOKEN`.
   - `POST` receipt of `messages` webhook payloads, verified via the
     `X-Hub-Signature-256` HMAC-SHA256 header against the app secret before
     any parsing — this is the Cloud API's substitute for Baileys' transport
     trust and must not be skipped.
   - Respond `200` immediately (Meta requires <5s ack), then normalise to
     `IncomingMessage` and hand off to the existing `MessageHandler` /
     `router.ts` path asynchronously — matches the "no changes to router/
     agent/storage" promise already made in ARCHITECTURE.md.
2. **Outbound** — `sendMessage` → `POST
   https://graph.facebook.com/v21.0/{PHONE_NUMBER_ID}/messages`. Free-form
   replies only work inside the 24h customer-service window opened by an
   inbound message; outside that window, only pre-approved message templates
   can be sent (relevant to `announce` / admin DMs — needs a
   template-vs-free-form branch, or an explicit "outside window" failure
   mode instead of silently dropping).
3. **Admin actions** — the Cloud API has no group/moderation concept (1:1 and
   business-initiated messaging only), so `adminCapabilities` stays empty as
   today's stub already declares; `performAdminAction` should return a clear
   "unsupported on Cloud API" rather than throwing an opaque error.
4. **Config** — `WHATSAPP_PROVIDER=cloud` plus `WHATSAPP_CLOUD_PHONE_NUMBER_ID`,
   `WHATSAPP_CLOUD_ACCESS_TOKEN`, `WHATSAPP_CLOUD_APP_SECRET`,
   `WHATSAPP_CLOUD_VERIFY_TOKEN` (extend `src/config.ts` zod schema; only
   required when the provider is selected).
5. Update `docs/DEPLOYMENT.md` with reverse-proxy/webhook exposure guidance
   and `docs/ARCHITECTURE.md`'s "Switching WhatsApp providers" section to
   mark this as implemented rather than an upgrade path.

## Alternatives considered

- **Stay on Baileys, do nothing** — keeps the ban risk; rejected, since it's
  the explicitly documented top risk and the adapter interface already exists
  to solve it.
- **Third-party BSP (Twilio, 360dialog, etc.)** — faster onboarding, better
  support, but adds a paid middleman and a second vendor dependency for a
  community project; the direct Meta Cloud API is free to use (message-based
  pricing only) and keeps the stack in-house. Worth revisiting if webhook
  hosting/ops becomes a burden.
- **Make Cloud API a drop-in default and delete Baileys** — premature; keep
  both behind `WHATSAPP_PROVIDER` so self-hosters without a Meta Business
  verification can still run Baileys, per the existing design.

## Security / privacy impact

- New public HTTPS surface (the webhook endpoint) — must be scoped tightly:
  verify-token check on GET, HMAC signature check on every POST before
  touching the body, and the endpoint must not leak into the same origin as
  anything else sensitive.
- No change to RBAC, memory, or outbound filtering — inbound normalises into
  the same `IncomingMessage` shape the router already gates and audits.
- Meta stores message metadata on their side per their own retention terms;
  worth a one-line mention in SECURITY.md's privacy section alongside the
  existing WhatsApp Baileys caveat.
- Access token and app secret are new secrets — must go through the same
  `.env`-only, git-ignored handling as existing tokens.

## Rough scope

- `src/platforms/whatsapp/cloudAdapter.ts` — full implementation (medium
  file, mirrors `baileysAdapter.ts` shape).
- `src/config.ts` — new env vars, conditionally required.
- Webhook route wiring (likely in `src/router.ts` or a small new
  `src/platforms/whatsapp/cloudWebhook.ts` + whatever the service uses to
  listen on HTTP — check if one already exists for health checks).
- `docs/DEPLOYMENT.md`, `docs/ARCHITECTURE.md`, `README.md` updates.
- Tests: webhook signature verification (valid/invalid/missing signature),
  verify-token handshake, inbound payload → `IncomingMessage` normalisation,
  24h-window outbound branching. This lands squarely in "WhatsApp wire
  helpers," which CLAUDE.md already calls out as security-invariant test
  territory.

## Acceptance criteria

- [ ] `WHATSAPP_PROVIDER=cloud` boots the adapter and passes webhook
      verification against a real (or mocked) Meta subscription.
- [ ] Unsigned or badly-signed webhook POSTs are rejected before any
      processing, with a test proving it.
- [ ] Inbound Cloud API messages reach the same router/agent/storage path as
      Baileys messages, unmodified.
- [ ] Outbound free-form replies work inside the 24h window; sends outside
      the window fail clearly (not silently) rather than attempting an
      unsupported free-form send.
- [ ] `performAdminAction` returns a clear unsupported-operation result
      instead of an unhandled throw.
- [ ] Docs updated to reflect Cloud API as a working option, not just a stub.

swampratnz:
Closes #4

## Summary

Replaces the `WhatsAppCloudAdapter` stub (`src/platforms/whatsapp/cloudAdapter.ts`) with a real implementation against the official WhatsApp Business Cloud API, giving self-hosters a ToS-compliant alternative to `BaileysAdapter` (which uses the unofficial linked-device protocol and risks a number ban).

- **Webhook intake** via a `node:http` server (no new dependency needed — Node 22's built-in `fetch`/`http`/`crypto` cover everything):
  - `GET` — verification handshake (`hub.mode`/`hub.verify_token`/`hub.challenge`) against `WHATSAPP_CLOUD_VERIFY_TOKEN`.
  - `POST` — verifies `X-Hub-Signature-256` (HMAC-SHA256 over the raw body) against `WHATSAPP_CLOUD_APP_SECRET` **before** parsing anything, then acks `200` immediately (Meta requires <5s) and normalises text messages into the same `IncomingMessage` shape Baileys produces, handed to the same router/agent path unmodified.
  - Pure parsing/verification logic lives in a new `src/platforms/whatsapp/cloudWire.ts` (mirrors the existing `wire.ts` pattern), kept free of HTTP/config imports so it's directly unit-testable.
- **Outbound** — `sendMessage`/`sendDirectMessage` POST to the Graph API (`v21.0`), routed through the same `filterOutbound` secret-redaction/code-policy path every other adapter uses. Free-form sends are only attempted within Meta's 24h customer-service window, tracked in-process from each sender's last inbound message; outside that window the send fails immediately with a clear error instead of attempting (and getting rejected by) an unsupported free-form message.
- **Admin actions** — the Cloud API has no group/moderation concept, so `adminCapabilities` only advertises `warn_user` (DM-based). `performAdminAction` throws a clear "unsupported" error for anything else, and the existing `moderate` tool already gates on `adminCapabilities.has(...)` before ever calling it, so an admin gets a clean "not supported on this platform" message rather than a throw.
- **Config** — added `WHATSAPP_CLOUD_APP_SECRET` (needed for webhook signature verification, previously missing from the schema) and removed the hard block on `WHATSAPP_PROVIDER=cloud`, replacing it with a `zod` refine requiring all four `WHATSAPP_CLOUD_*` vars when `cloud` is selected. No changes to `router.ts`, `agent/*`, or `storage/*` — purely additive behind the existing `PlatformAdapter` interface, matching the design already documented in `docs/ARCHITECTURE.md`.
- **Docs** — updated `README.md`, `docs/ARCHITECTURE.md` ("Switching WhatsApp providers"), `docs/SECURITY.md` (new webhook-security subsection), `docs/DEPLOYMENT.md` (new "Configure the WhatsApp Cloud API instead of Baileys" section + troubleshooting entries), and `.env.example`.

## Security impact

- New public HTTP surface (the webhook listener) is scoped tightly: every `POST` is rejected with `401` unless its HMAC-SHA256 signature verifies against `WHATSAPP_CLOUD_APP_SECRET`, checked before the body is ever parsed. The GET handshake requires an exact `WHATSAPP_CLOUD_VERIFY_TOKEN` match or returns `403`.
- No RBAC, memory, or outbound-filtering changes — inbound Cloud API messages normalise into the same `IncomingMessage` the router already gates/audits/rate-limits; outbound sends go through the same `filterOutbound` (secret redaction + code policy) as Baileys/Discord.
- `WHATSAPP_CLOUD_ACCESS_TOKEN` and `WHATSAPP_CLOUD_APP_SECRET` are new secrets, handled the same `.env`-only, git-ignored way as existing tokens (documented in `docs/SECURITY.md`).
- Admin data scoping is unaffected: `conversationsForUser` returns only the caller's own 1:1 (no groups exist on this API), so admin actions can't be scoped to conversations they don't participate in.

## Verification

- `npm test` — 50/50 passing, including a new `tests/whatsappCloudWire.test.ts` covering: valid/invalid/missing webhook signatures (wrong signature, wrong secret, malformed prefix, wrong-length hex — all `SECURITY:`-prefixed), the verification handshake, and payload normalisation (well-formed text messages, non-text/status payloads ignored, malformed payloads yield `[]`).
- `npm run build` — clean.
- Manual smoke test of the live HTTP server (via `dist/`, not just unit tests): verified the `GET` handshake returns `200` + echoes the challenge on a correct token and `403` on a wrong one; a signed `POST` webhook is normalised and delivered to the message handler correctly; an unsigned/badly-signed `POST` is rejected `401` before parsing; `sendMessage` to a recently-messaged user correctly attempts the real Graph API call; `sendMessage` to a user outside the 24h window fails immediately with a clear error instead of attempting the send.
- Not exercised against a live Meta Business account/real webhook subscription (would require Meta app credentials); the above smoke test exercises the full request/response path against a mocked payload with a real HMAC signature.

---

### DIFF
--- a/src/platforms/whatsapp/cloudAdapter.ts
+++ b/src/platforms/whatsapp/cloudAdapter.ts
@@ -1,52 +1,248 @@
+import { createServer, type IncomingMessage as HttpRequest, type Server, type ServerResponse } from 'node:http';
+import { config } from '../../config.js';
+import { logger } from '../../logger.js';
+import { filterOutbound } from '../../agent/outbound.js';
+import { runtimeSecrets } from '../../agent/secrets.js';
+import { getCodeAnswersPolicy } from '../../storage/policies.js';
+import { extractMessages, parseVerificationRequest, verifySignature, type CloudInboundMessage } from './cloudWire.js';
 import type {
   AdminAction,
+  IncomingMessage,
   MessageHandler,
   OutgoingMessage,
   PlatformAdapter,
 } from '../types.js';
 
+const GRAPH_API_VERSION = 'v21.0';
+/** Meta webhook payloads are small JSON; this bounds memory use per request. */
+const MAX_BODY_BYTES = 1_000_000;
+/** Free-form replies are only allowed within Meta's 24h customer-service window. */
+const CUSTOMER_SERVICE_WINDOW_MS = 24 * 60 * 60_000;
+
 /**
- * Stub for the official WhatsApp Business Cloud API (Meta).
+ * WhatsApp via the official Meta Business Cloud API. ToS-compliant
+ * alternative to {@link BaileysAdapter}: no linked-device session to ban,
+ * webhook-driven, 1:1 messaging only (no group/moderation surface).
  *
- * This is intentionally unimplemented: it documents the seam so the project
- * can move off Baileys to the official, ToS-compliant API without touching the
- * agent core or router. To implement:
- *  1. Stand up an HTTPS webhook endpoint (verify with WHATSAPP_CLOUD_VERIFY_TOKEN).
- *  2. On inbound `messages` webhooks, normalise to IncomingMessage and call the handler.
- *  3. sendMessage -> POST https://graph.facebook.com/v21.0/{PHONE_NUMBER_ID}/messages.
- *  4. Map admin actions (the Cloud API has no group moderation; most will be unsupported).
- *
- * See docs/ARCHITECTURE.md "Switching WhatsApp providers".
+ * Requires WHATSAPP_CLOUD_PHONE_NUMBER_ID, _ACCESS_TOKEN, _VERIFY_TOKEN, and
+ * _APP_SECRET (see config.ts). Set WHATSAPP_PROVIDER=cloud to select it.
  */
 export class WhatsAppCloudAdapter implements PlatformAdapter {
   readonly platform = 'whatsapp' as const;
-  readonly adminCapabilities = new Set<string>([]);
+  readonly adminCapabilities = new Set(['warn_user']);
 
-  onMessage(_handler: MessageHandler): void {
-    throw new Error('WhatsAppCloudAdapter is not implemented yet. Set WHATSAPP_PROVIDER=baileys.');
+  private handler: MessageHandler | null = null;
+  private server: Server | null = null;
+  /**
+   * Last time each user sent an inbound message, tracked in-process to
+   * enforce Meta's 24h free-form messaging window. Resets on restart — a
+   * known limitation of not persisting this; see docs/DEPLOYMENT.md.
+   */
+  private readonly lastInboundAt = new Map<string, number>();
+
+  onMessage(handler: MessageHandler): void {
+    this.handler = handler;
   }
 
   async start(): Promise<void> {
-    throw new Error('WhatsAppCloudAdapter is not implemented yet. Set WHATSAPP_PROVIDER=baileys.');
+    const { phoneNumberId, accessToken, verifyToken, appSecret, webhookPort } = config.whatsapp.cloud;
+    if (!phoneNumberId || !accessToken || !verifyToken || !appSecret) {
+      throw new Error(
+        'WhatsAppCloudAdapter requires WHATSAPP_CLOUD_PHONE_NUMBER_ID, _ACCESS_TOKEN, _VERIFY_TOKEN, and _APP_SECRET',
+      );
+    }
+
+    await new Promise<void>((resolve, reject) => {
+      const server = createServer((req, res) => {
+        this.handleRequest(req, res).catch((err) => {
+          logger.error({ err }, 'WhatsApp Cloud webhook handler failed');
+          if (!res.headersSent) res.writeHead(500).end();
+        });
+      });
+      server.once('error', reject);
+      server.listen(webhookPort, () => {
+        server.off('error', reject);
+        resolve();
+      });
+      this.server = server;
+    });
+    logger.info({ port: webhookPort }, 'WhatsApp Cloud webhook listening');
   }
 
   async stop(): Promise<void> {
-    /* nothing to do */
+    const server = this.server;
+    this.server = null;
+    if (!server) return;
+    await new Promise<void>((resolve) => server.close(() => resolve()));
   }
 
-  async sendMessage(_out: OutgoingMessage): Promise<void> {
-    throw new Error('WhatsAppCloudAdapter.sendMessage not implemented.');
+  private async handleRequest(req: HttpRequest, res: ServerResponse): Promise<void> {
+    const url = new URL(req.url ?? '/', 'http://localhost');
+    if (req.method === 'GET') {
+      this.handleVerification(url, res);
+      return;
+    }
+    if (req.method === 'POST') {
+      await this.handleWebhook(req, res);
+      return;
+    }
+    res.writeHead(405).end();
   }
 
-  async sendDirectMessage(_userId: string, _text: string): Promise<void> {
-    throw new Error('WhatsAppCloudAdapter.sendDirectMessage not implemented.');
+  private handleVerification(url: URL, res: ServerResponse): void {
+    const verification = parseVerificationRequest(url);
+    const { verifyToken } = config.whatsapp.cloud;
+    if (verification && verification.mode === 'subscribe' && verification.token === verifyToken) {
+      res.writeHead(200, { 'Content-Type': 'text/plain' }).end(verification.challenge);
+      return;
+    }
+    logger.warn('Rejected WhatsApp Cloud webhook verification request');
+    res.writeHead(403).end();
   }
 
-  async conversationsForUser(_userId: string): Promise<string[]> {
-    throw new Error('WhatsAppCloudAdapter.conversationsForUser not implemented.');
+  private async handleWebhook(req: HttpRequest, res: ServerResponse): Promise<void> {
+    let rawBody: Buffer;
+    try {
+      rawBody = await readBody(req);
+    } catch (err) {
+      logger.warn({ err }, 'WhatsApp Cloud webhook: failed to read body');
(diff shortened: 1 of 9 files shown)


src/platforms/whatsapp/cloudAdapter.ts
swampratnz, line 1: `WHATSAPP_ALLOWED_JIDS` is reused here (`config.whatsapp.allowedJids.includes(msg.from)`), but on the Baileys adapter that list holds full JIDs (`64211234567@s.whatsapp.net`, `...@g.us`), whereas `msg.from` here is a bare phone-number digit string (per `CloudInboundMessage.from`'s own doc comment: "E.164 digits, no '+'"). The formats never match.

Net effect: if an operator sets `WHATSAPP_ALLOWED_JIDS` (a security-relevant allowlist) expecting it to also restrict the Cloud API adapter, every inbound Cloud message will be silently dropped instead — the adapter goes completely non-functional with no error logged to explain why. It fails closed rather than open, so it's not an auth bypass, but it's a correctness bug in a security control and will be confusing to debug (nothing in the PR docs calls out that the two adapters expect different formats for this var).

Worth either: normalising `msg.from` to the same JID shape before the allowlist check, or documenting/enforcing that Cloud API operators must list bare digits in `WHATSAPP_ALLOWED_JIDS` (and ideally validating/rejecting mixed-format entries at startup).

---

swampratnz, line 1: Minor/non-blocking: `lastInboundAt` is an unbounded `Map` keyed by sender that's never pruned, so it grows for the life of the process in a busy community. Not urgent (bounded by community size, resets on restart), but a periodic sweep of stale entries (older than the 24h window) would keep it tidy for long-running deployments.

---

swampratnz (commented):
Solid implementation overall — signature verification is correct (HMAC-SHA256, timing-safe compare, verified before body parsing), webhook ack-then-process pattern respects Meta's 5s limit, outbound sends go through the same `filterOutbound` path as other adapters, admin action surface is honestly scoped to what the Cloud API actually supports, and the new `cloudWire.ts` has good unit test coverage of the security-relevant paths (signature edge cases, malformed payloads). CI (typecheck/test/build) is green.

One correctness bug worth fixing before merge: the `WHATSAPP_ALLOWED_JIDS` allowlist check reuses a config field whose format (full JIDs) doesn't match what this adapter compares it against (bare phone digits) — see inline comment. It fails closed (drops messages rather than admitting unauthorized ones), so not a security hole, but it silently breaks the allowlist feature for Cloud API operators. Also left a minor non-blocking note about unbounded map growth.

Not marking this "ready for human merge" until the allowlist format mismatch is addressed.

---

src/platforms/whatsapp/cloudAdapter.ts
swampratnz: Fixed in 6f878e6 — added `isAllowedSender()` (cloudWire.ts) which normalises by stripping the `@...` suffix, so the shared `WHATSAPP_ALLOWED_JIDS` list works for either full-JID (Baileys) or bare-digit (Cloud) entries. Also documented the format compatibility in ARCHITECTURE.md and added test coverage for both formats.

---

swampratnz: Fixed in 6f878e6 — added a periodic sweep (hourly) that drops `lastInboundAt` entries older than the 24h window.

---

swampratnz:
Fix commit (6f878e6) resolves both review points:

- `isAllowedSender()` in `cloudWire.ts` normalizes by matching the part before `@`, so `WHATSAPP_ALLOWED_JIDS` entries work whether they're bare digits or full Baileys-style JIDs — covered by new tests, including one explicitly exercising the shared-config case.
- `lastInboundAt` now gets swept hourly for entries older than the 24h window, addressing the unbounded-growth nit.

Docs (`ARCHITECTURE.md`) were updated to explain the shared allowlist format. CI is green.

LGTM — ready for human merge.

---

PR #8, opened 2026-07-02
Add knowledge curation tools: list, update, delete

### LINKED ISSUE #5: Proposal: knowledge curation tools — list, update, retire entries
## Problem

`knowledge` (`schema.sql`) and `save_knowledge` (`src/agent/tools.ts`) let an
admin **add** a durable fact/FAQ via chat, and the agent can **search** it
(`knowledge_search`). That's the entire lifecycle currently implemented in
`src/storage/repository.ts` — insert, semantic search, and a cascade-delete
that only fires when a user's data is purged (`purgeUserData`). There is no
tool to:

- list existing knowledge entries (by scope, recency, or author),
- correct one that's gone stale or was entered wrong, or
- retire/delete a specific entry that's no longer accurate,

without an admin going around the bot entirely and running raw SQL against
production Postgres. For a community FAQ/rules store that's meant to answer
questions accurately, "the only fix for a wrong answer is a database console"
is a real operational gap — and it also means the agent itself has no way to
"un-learn" something an admin flags as outdated mid-conversation, since
`save_knowledge` can only add, never supersede.

## Proposed approach

Add three admin-gated tools to `src/agent/tools.ts`, backed by new
`src/storage/repository.ts` functions:

1. **`list_knowledge`** — paginated list filtered by `scope` and/or a text/
   date filter (e.g. `ORDER BY updated_at DESC LIMIT/OFFSET`), returning
   `id`, `title`, `content` (maybe truncated), `scope`, `created_by_role`,
   `updated_at`. Lets an admin actually see what's stored before deciding to
   fix anything — today there's no read path at all besides semantic search,
   which won't surface everything (e.g. it won't reliably return an entry
   whose embedding is a poor match for any query the admin thinks to type).
2. **`update_knowledge`** — takes an `id` plus new `title`/`content`/`scope`,
   re-embeds the content (mirrors the embedding-on-insert path in
   `repository.ts:300-303`), and updates `updated_at` (the schema already has
   an `updated_at` trigger — `schema.sql:79-81` — that's currently unused by
   application code since nothing ever updates a row).
3. **`delete_knowledge`** — takes an `id`, deletes the row. Should probably
   route through the existing `audited()` wrapper used by `moderate`/
   `announce` (`src/agent/tools.ts`) since it's a destructive admin action —
   whether it needs the same out-of-band CONFIRM gate as `moderate`/`announce`
   is the open question for adversarial review (arguably yes: it's
   irreversible and affects what the bot tells every future asker).

All three `assertAtLeast(caller.role, 'admin', ...)`, matching `save_knowledge`.
No RBAC tier or scoping model changes — same admin/super_admin split as
existing knowledge operations.

## Alternatives considered

- **Do nothing; require direct DB access for corrections** — the status quo;
  rejected because it defeats the audit trail (`admin_audit`) that every
  other privileged mutation goes through, and puts a Postgres credential in
  the hands of anyone who needs to fix a typo in an FAQ.
- **Soft-delete/versioning (keep history of edits)** — more correct for an
  auditable knowledge base, but adds schema migration scope (a `knowledge_history`
  table or `deleted_at` column) beyond what a first pass needs. Worth a
  follow-up if the community wants an edit history view; not blocking for v1.
- **Let `save_knowledge` upsert by title instead of adding new tools** —
  simpler surface, but silently overwriting by title match is fragile (title
  collisions, no explicit "which one did I mean") and loses the list/browse
  capability admins actually need to find what to fix.

## Security / privacy impact

- All three tools are `admin`+-gated only, same as `save_knowledge` — no new
  role or tool-gating surface.
- `delete_knowledge`/`update_knowledge` should go through `audited()` so
  super admins get the same DM alert + `admin_audit` row as other privileged
  mutations (CLAUDE.md: "everything privileged is audited + alerted").
- `list_knowledge` returning content back to the model is no different in
  kind from `knowledge_search` already doing so — no new data exposure, since
  admins can already read any knowledge entry via search. It does make full
  enumeration easier than semantic search does, which is the point.
- No user-supplied message content is trusted as instructions here — inputs
  are `id`/`scope` filters from an already-role-checked caller.

## Rough scope

- `src/storage/repository.ts` — `listKnowledge`, `updateKnowledge`,
  `deleteKnowledge` functions.
- `src/agent/tools.ts` — `list_knowledge`, `update_knowledge`,
  `delete_knowledge` tool defs, wired into the admin tool set.
- Tests extending existing knowledge/RBAC test coverage: role gating (member
  can't call these), update re-embeds and bumps `updated_at`, delete actually
  removes the row and doesn't affect unrelated entries, audit trail is
  written for update/delete.
- No schema migration needed — `knowledge` table already has everything
  required (`updated_at` trigger already exists and is currently dead code).

## Acceptance criteria

- [ ] Admin can list knowledge entries (optionally filtered by scope) without
      direct DB access.
- [ ] Admin can update an existing entry's title/content/scope; the stored
      embedding reflects the new content and `updated_at` changes.
- [ ] Admin can delete a specific entry by id; `knowledge_search` no longer
      returns it afterward.
- [ ] Member-tier callers get these tools denied (not offered / `assertAtLeast`
      throws), consistent with `save_knowledge`.
- [ ] Update/delete are recorded via the existing audit path
      (`admin_audit` + super-admin DM alert).

swampratnz:
Closes #5

## Summary

Rounds out the knowledge lifecycle. Previously `save_knowledge` could only *add* an entry and `knowledge_search` could semantically search it — there was no way to browse, correct, or retire an entry without going around the bot and running raw SQL against production Postgres.

- **`listKnowledge` / `updateKnowledge` / `deleteKnowledge`** (`src/storage/repository.ts`) — new repository functions:
  - `listKnowledge({ scope?, limit?, offset? })` — paginated, `ORDER BY updated_at DESC`, optional scope filter. Lets an admin see everything stored, including entries semantic search might not surface for a given query.
  - `updateKnowledge({ id, title?, content?, scope? })` — partial update (each field optional, unspecified fields keep their current value), re-embeds using the same `title\ncontent` convention as `saveKnowledge`, and relies on the schema's existing `updated_at` trigger (previously dead code — nothing ever updated a row before this).
  - `deleteKnowledge(id)` — straightforward delete by id; returns `false` if nothing matched so callers can report a clear "not found" instead of a silent no-op.
- **Three new tools** (`src/agent/tools.ts`): `list_knowledge`, `update_knowledge`, `delete_knowledge`. All `assertAtLeast(caller.role, 'admin', ...)`, matching `save_knowledge`. `update_knowledge` and `delete_knowledge` route through the existing `audited()` wrapper, so they get the same `admin_audit` row + super-admin DM alert as every other privileged mutation. `delete_knowledge` is additionally CONFIRM-gated via `requireConfirm` (irreversible and affects what the bot tells every future asker — same reasoning as `delete_message`/`kick_user`), with `assertAtLeast` checked eagerly (before queuing), matching the `grant_admin`/`purge_user_data` pattern.
- **RBAC** (`src/auth/rbac.ts`): added the three new tool names to `ADMIN_TOOLS`. No new tier or scoping model — `toolsForRole`'s structural gating and the existing RBAC tests (which iterate `ADMIN_TOOLS`) cover them automatically.
- **Docs**: updated `docs/ARCHITECTURE.md`'s memory section and RBAC capability table.
- No schema migration — the `knowledge` table already had everything needed (the `updated_at` trigger existed but was unused).

## Security impact

- All three tools are admin+-gated only, identical surface to `save_knowledge` — no new role or tool-gating model.
- `update_knowledge`/`delete_knowledge` are audited (DB row + super-admin DM), matching CLAUDE.md's "everything privileged is audited + alerted" invariant.
- `delete_knowledge` requires an out-of-band CONFIRM reply, handled deterministically by the router/pendingActions machinery — never executed by the model directly.
- `list_knowledge` returning content to the model is no wider than `knowledge_search` already returning content — it just makes full enumeration easier (the intended change), not new data exposure.
- No user-supplied message content is trusted as instructions — inputs are `id`/`scope` filters from an already role-checked caller, and recalled content returned to the model goes through the existing `untrusted()` framing.

## Verification

- `npm test` — 53/53 passing (no regressions; RBAC tests iterate `ADMIN_TOOLS`/`SUPER_ADMIN_TOOLS` so the three new tool names are automatically covered by the existing gating tests — members/guests never get them, admins do, super-admins do).
- `npm run build` — clean.
- **Exercised against a local Postgres 16 + pgvector** (per CLAUDE.md): ran `npm run migrate` against a fresh local DB, then manually drove `saveKnowledge` → `listKnowledge` → `updateKnowledge` → `deleteKnowledge` → `listKnowledge` end-to-end via the built `dist/`, confirming:
  - `listKnowledge` returns entries newest-first and the `scope` filter only returns matching rows.
  - `updateKnowledge` preserves unspecified fields (title unchanged when only `content` is passed), the content actually changes, `updated_at` is bumped, and the new content is found by `searchKnowledge` (proving re-embedding happened).
  - `updateKnowledge`/`deleteKnowledge` on a nonexistent id return `false` rather than throwing.
  - Deleting one entry removes it from `listKnowledge` while leaving an unrelated entry untouched.
- No automated DB-integration tests were added to `tests/*.test.ts` — CI (`ci.yml`) has no Postgres service and no existing test in the repo touches the DB (all existing tests exercise pure helpers or in-memory RBAC/pendingActions logic); adding one here would break `npm test` for anyone without a local Postgres. This matches the existing repo convention.

---

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -67,7 +67,10 @@ memory**:
    (never the system prompt — see SECURITY.md on prompt injection).
 3. The `remember_search` / `knowledge_search` tools let the model query memory
    on demand mid-turn. Cross-conversation search is admin-only.
-4. Admins can promote durable facts into `knowledge` via `save_knowledge`.
+4. Admins can promote durable facts into `knowledge` via `save_knowledge`, and
+   curate existing entries with `list_knowledge` (browse by scope),
+   `update_knowledge` (correct + re-embed), and `delete_knowledge` (retire,
+   CONFIRM-gated).
 
 Conversation continuity uses the Agent SDK's session resume: the Claude
 `session_id` for each `(platform, conversation)` is stored in `sessions` and
@@ -102,6 +105,7 @@ and every privileged action is audited and alerted to super admins by DM.
 | Search memory (own conversation), knowledge, `forget_me` | ❌ | ✅ | ✅ | ✅ |
 | Memory/history across conversations | ❌ | ❌ | ✅ *their conversations* | ✅ all |
 | `moderate` / `announce` | ❌ | ❌ | ✅ *their conversations*, confirm-gated | ✅ anywhere |
+| `save_knowledge` / `list_knowledge` / `update_knowledge` / `delete_knowledge` | ❌ | ❌ | ✅, delete confirm-gated | ✅ |
 | `add_member` / `remove_member` | ❌ | ❌ | ✅ (member tier only) | ✅ |
 | Web search & summarise (`WebSearch`; `WebFetch` never) | ❌ | ❌ | ✅ | ✅ |
 | `grant_admin` / `revoke_admin`, `purge_user_data`, `audit_view`, `usage_stats`, `pause_bot`, `set_policy` | ❌ | ❌ | ❌ | ✅ |
--- a/src/agent/tools.ts
+++ b/src/agent/tools.ts
@@ -5,9 +5,11 @@ import { assertAtLeast, type CallerContext } from '../auth/rbac.js';
 import { isSuperAdmin, superAdminIds } from '../auth/roles.js';
 import { logger } from '../logger.js';
 import {
+  deleteKnowledge,
   demoteAdmin,
   isKnownConversation,
   isKnownUser,
+  listKnowledge,
   purgeUserData,
   recentAuditEntries,
   recordAdminAction,
@@ -15,6 +17,7 @@ import {
   saveKnowledge,
   searchKnowledge,
   searchMemory,
+  updateKnowledge,
   upsertMember,
   usageStats,
   userMessages,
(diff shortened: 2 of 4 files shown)


swampratnz:
Reviewed. This is clean:

- All three tools are admin-gated (`assertAtLeast(caller.role, 'admin', ...)`), matching `save_knowledge`'s existing surface — no new tier/scoping model.
- `update_knowledge`/`delete_knowledge` go through `audited()` (DB row + super-admin DM), and `delete_knowledge` is CONFIRM-gated with `assertAtLeast` checked eagerly before queuing — consistent with the `grant_admin`/`purge_user_data` pattern (minTier re-checked at confirm time via the router).
- SQL is fully parameterized in `repository.ts` (`listKnowledge`/`updateKnowledge`/`deleteKnowledge`) — no injection surface; the scope filter's column is fixed, only the value is bound.
- RBAC test coverage is inherited automatically since the new tool names were added to `ADMIN_TOOLS`, which existing tests iterate over.
- CI green.

One minor, non-blocking note: in `updateKnowledge` (`src/storage/repository.ts`), if `embed()` throws, the catch logs a warning and `embedding` stays `null`, so `COALESCE($5, embedding)` keeps the *old* embedding while `title`/`content` are still updated unconditionally. That leaves `searchKnowledge` (embedding-based) results briefly stale relative to the new text after a failed re-embed — not a security issue, just a data-consistency edge case worth a code comment or a retry/flag if it turns out to matter in practice.

LGTM — ready for human merge.

---

PR #10, opened 2026-07-02
Add proactive onboarding: Discord welcome message + access-request queue

swampratnz:
Closes #7

## Summary

Onboarding in gated mode (the default) was purely reactive: new Discord joiners got no welcome at all, and every guest ping to the bot was handled independently and forgotten — admins had no visibility into who'd actually asked to be added short of an out-of-band ping.

Two additive, low-risk pieces:

1. **Welcome message on join** (`src/platforms/discord/adapter.ts`) — a `GuildMemberAdd` listener sends a static, non-agent DM (no LLM call, same cost profile as the existing `GATED_NOTICE`) explaining what the bot is and that they should ask an admin. Off unless `DISCORD_WELCOME_ENABLED=true`, so existing deployments are unaffected. DM-first; falls back to posting in `DISCORD_WELCOME_CHANNEL_ID` if the member's DMs are closed. Discord-only — WhatsApp has no equivalent "join" event, noted explicitly rather than faked.
2. **Pending-access queue** — a new `access_requests` table (`src/storage/schema.sql`) storing `platform`, `user_id`, `user_name`, `first_requested_at`, `last_requested_at`, `request_count`. `router.ts` upserts a row whenever a gated guest addresses the bot, alongside (not instead of) the existing rate-limited `GATED_NOTICE` — deliberately **never** their message content, preserving the existing "gated guest content is not stored" invariant. New `list_access_requests` tool (`src/agent/tools.ts`, admin-gated) lets an admin ask "who's waiting to be added" and get a real list. `add_member` clears a user's row once they're actioned.

## Security impact

- `access_requests` stores identity + request metadata only, **no message content** — verified in the manual smoke test below (no `content`/`text` field on the returned rows).
- `list_access_requests` is `admin`+-gated via `assertAtLeast`, same pattern as `usage_stats`/`audit_view`. Per the proposal's explicit call-out: **unlike other admin data-access tools, this one is not scoped to "conversations the admin participates in"** — a guest's access request isn't tied to a conversation the admin is a member of the same way (they're pre-membership by definition), so there's nothing to scope against. No new data exposure beyond what `add_member` already implies admins can see (that a user id/name exists and wants access).
- The welcome DM is a static, non-agent message — no prompt-injection surface (never calls the model), no cost.
- Row growth is bounded: upsert-on-repeat-request (not one row per ping), and `add_member` clears resolved rows.

## Verification

- `npm run typecheck` — clean.
[...]

PR #12, opened 2026-07-02
Add age-based retention purge for raw interactions

### LINKED ISSUE #9: Proposal: automated retention policy for raw interactions (age-based purge)
## Problem

`docs/SECURITY.md` names this as an open item twice and it's still unbuilt:

- §6 Data protection: "Consider a retention policy (e.g. purge raw
  `interactions` older than N months while keeping aggregate `knowledge`)."
- Operational checklist: "- [ ] A retention/deletion policy is defined."

Today the only deletion paths are **per-user** and **manual**: `forget_me`
(member, self-service) and `purge_user_data` (super admin) in
`src/agent/tools.ts`, both driven by `src/storage/repository.ts`. There is no
mechanism that reduces the standing footprint of `interactions` over time —
every message ever recorded (member+ traffic; see `router.ts`) stays in
Postgres indefinitely unless an individual is explicitly purged. For a bot
whose own docs invoke the **NZ Privacy Act 2020** and tell operators to
"define a retention/deletion policy," shipping with no age-based purge at all
is a real gap between the stated design and the implementation — every
self-hoster currently has to hand-roll this themselves or skip it.

## Proposed approach

Add a scheduled, config-driven purge of **raw** `interactions` content,
mirroring the "purge raw messages, keep aggregate knowledge" split SECURITY.md
already recommends:

1. **Config** — `INTERACTION_RETENTION_DAYS` (optional; unset/`0` = retention
   disabled, preserving current behaviour for existing deployments — no
   silent behaviour change on upgrade).
2. **Repository function** — `purgeOldInteractions(days)` in
   `src/storage/repository.ts`: `DELETE FROM interactions WHERE created_at <
   now() - interval '$days days'`. Deliberately **does not** touch
   `knowledge` (curated facts are meant to be durable — that's the point of
   promoting something into knowledge via `save_knowledge`), `sessions`
   (needed for active conversation continuity — should respect
   `SESSION_MAX_TURNS`/`_AGE_HOURS` separately), or `admin_audit`
   (accountability trail is explicitly retained deliberately per
   `SECURITY.md` residual risks, same reasoning as `forget_me`/`purge_user_data`
   already keeping it).
3. **Scheduling** — a lightweight `setInterval` in the service entrypoint
   (daily cadence is plenty; this doesn't need a cron dependency) rather than
   introducing a job-scheduling library — matches the project's existing
   "single long-running Node service" shape (`docs/ARCHITECTURE.md`).
4. **Observability** — log the row count purged each run (info level) so
   operators can see it's working without needing DB access; optionally
   surface last-purge time/count via the existing `usage_stats` /
   `audit_view` super-admin tools rather than adding a new one.

## Alternatives considered

- **Leave it as an operator responsibility (status quo)** — technically
  matches "self-hosted, configure it yourself," but the project already
  ships opinionated defaults for everything else security/privacy-related
  (secret redaction, outbound filtering, RBAC) and explicitly calls this one
  out as unfinished in its own checklist. Rejected as inconsistent with the
  rest of the project's posture.
- **External cron + a purge CLI script** (`npm run purge-old`) instead of an
  in-process timer — simpler to reason about, no timer inside the long-running
  service, and lets ops choose their own scheduler. Real alternative; the
  trade-off is it requires operators to actually wire up a system cron job
  (another manual step, same failure mode as today), whereas an opt-in env
  var "just works" once set. Worth raising in adversarial review — could ship
  both (script for one-off/manual runs, optional timer for zero-touch).
- **Anonymise instead of delete** (strip identifying fields, keep message
  text for trend analysis) — more complex, and the existing `usage_stats`
  already aggregates without needing raw retained content; plain delete is
  simpler and matches what `forget_me` already does per-user.
- **Soft-delete with a grace period** — adds complexity (extra column, a
  second sweep) for a benefit (undo window) nothing else in the deletion
  paths currently offers; `forget_me`/`purge_user_data` are already
  immediate, hard deletes, so staying consistent with that precedent.

## Security / privacy impact

- Net **reduces** stored PII footprint over time — directly addresses the
  privacy gap the project's own docs flag, no new exposure.
- No new RBAC surface if scheduling is a plain timer with no user-facing
  tool; if a manual-trigger tool is added instead, it should be
  `super_admin`-gated and audited like every other privileged mutation
  (`assertAtLeast` + `admin_audit`, per CLAUDE.md's security posture).
- Must not purge rows still referenced by anything relying on history within
  the retention window (e.g. memory recall for an active, ongoing
  conversation) — retention period should default conservatively (e.g. much
  longer than `SESSION_MAX_TURNS_AGE_HOURS`) so it doesn't silently degrade
  the "remembers past discussions" feature for active users; this needs an
  explicit minimum-days guard or a documented recommendation, not just a raw
  int.
- Purge query itself is a fixed, parameterised interval delete — no
  user-controllable input, no injection surface.

## Rough scope

- `src/config.ts` — `INTERACTION_RETENTION_DAYS` (optional, zod).
- `src/storage/repository.ts` — `purgeOldInteractions(days)`.
- Service entrypoint (wherever `router`/adapters are started) — daily timer
  wired up only when the env var is set.
- `docs/SECURITY.md` — mark the two open items done, document the default/
  recommended value and the "knowledge/audit/sessions are untouched" scope.
- `docs/DEPLOYMENT.md` — mention the env var in setup.
- Tests: purge deletes rows older than the cutoff and leaves newer ones;
  `knowledge`/`admin_audit`/`sessions` are unaffected; disabled (unset/`0`)
  is a no-op, preserving current behaviour.

## Acceptance criteria

- [ ] `INTERACTION_RETENTION_DAYS` unset (default) changes nothing about
      current behaviour.
- [ ] Set to N, interactions older than N days are removed on the scheduled
      run; newer rows, `knowledge`, `admin_audit`, and `sessions` are
      untouched.
- [ ] Purge activity is logged (count + timestamp) for operator visibility.
- [ ] `docs/SECURITY.md`'s two open retention-policy checklist/TODO items are
      resolved by this change.

swampratnz:
Closes #9

## Summary

`docs/SECURITY.md` flagged this as an open item twice ("consider a retention policy" / "a retention/deletion policy is defined" unchecked). Raw `interactions` had no age-based cleanup — only the per-user `forget_me`/`purge_user_data` paths existed, so every deployment's raw message footprint grew indefinitely unless an individual explicitly asked to be purged.

- **`src/config.ts`** — new `INTERACTION_RETENTION_DAYS` (optional, default `0` = disabled, so upgrading an existing deployment changes nothing unless explicitly configured). Enforced at startup via a `zod` refine: must be `0` or **at least 7 days** — a hard floor so a low value can't silently gut memory recall for users still mid-conversation (`SESSION_MAX_AGE_HOURS` defaults to 24h; 7 days gives real headroom above that).
- **`src/storage/repository.ts`** — `purgeOldInteractions(days)`: a single parameterised `DELETE FROM interactions WHERE created_at < now() - interval`. Deliberately never touches `knowledge` (curated, durable facts — the whole point of promoting something via `save_knowledge`), `admin_audit` (accountability trail, retained deliberately per existing SECURITY.md residual-risk notes), or `sessions` (governed separately by `SESSION_MAX_TURNS`/`_AGE_HOURS`).
- **`src/index.ts`** — a lightweight in-process daily timer (`setInterval`, `.unref()`'d, cleared on graceful shutdown), matching the project's existing "single long-running Node service" shape rather than adding a job-scheduling dependency. Runs once immediately at startup (so operators see it working without waiting a day) and logs the purged row count each run at `info` level. No timer/purge activity at all when the env var is unset — this is purely additive.
- **Docs** — `docs/SECURITY.md`'s two open retention-policy items are resolved (§6 Data protection, the operational checklist), and `docs/DEPLOYMENT.md`/`.env.example` document the new var.

Went with the in-process-timer approach over the "external cron + purge CLI script" alternative raised in the proposal: it's zero-touch once the env var is set (no separate system cron to wire up), consistent with how `SESSION_MAX_TURNS`/session hygiene and the router's own sweep timer already work in this codebase.

## Security impact

- Net **reduces** the stored PII footprint over time — directly addresses the privacy gap the project's own docs flagged; no new data exposure.
- No new RBAC surface — this is a config-driven background timer with no user-facing tool, so nothing to gate with `assertAtLeast`/audit (unlike `forget_me`/`purge_user_data`, which remain the on-demand, audited paths).
- The purge query is a single fixed, parameterised interval delete — no user-controllable input, no injection surface.
- The 7-day minimum floor is enforced at config-parse time (fails fast at startup with a readable error), not left as an unchecked raw int, addressing the proposal's explicit call-out that this needed a guard.

## Verification

- `npm test` — 53/53 passing, no regressions.
- `npm run build` — clean.
- **Exercised against a local Postgres 16 + pgvector** (per CLAUDE.md):
  - Confirmed the config guard: `INTERACTION_RETENTION_DAYS=3` fails startup with `INTERACTION_RETENTION_DAYS must be 0 (disabled) or at least 7`; unset (default `0`) and `30` both parse cleanly.
  - Manually drove `purgeOldInteractions` via the built `dist/`: inserted one recent interaction and one backdated 100-days-old interaction plus a 100-day-old `knowledge` entry, ran `purgeOldInteractions(30)`, and confirmed exactly the old interaction was deleted, the recent one survived, and the `knowledge` row was completely untouched despite being older than the retention window. Also confirmed a purge call with a far-future cutoff deletes nothing (no false positives).
- Did not exercise the `setInterval`-driven daily scheduling itself end-to-end (would require running the full service for 24h+); the scheduling logic is a thin wrapper calling the already-verified `purgeOldInteractions`, matching the existing router sweep-timer pattern (`router.ts`) already running unverified-by-test in production today.
- No automated DB-integration test added to `tests/*.test.ts` — CI has no Postgres service and no existing test in the repo touches the DB; matches the convention established in the two prior PRs (#8, #10) for this same reason.

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -66,6 +66,9 @@ DAILY_REPLY_LIMIT_PER_USER=50
 # Start a fresh Claude session past either cap (context/injection hygiene).
 SESSION_MAX_TURNS=30
 SESSION_MAX_AGE_HOURS=24
+# Age-based purge of raw `interactions` (privacy/retention). Unset/0 = disabled.
+# Must be 0 or >= 7 if set. knowledge/admin_audit/sessions are never touched.
+INTERACTION_RETENTION_DAYS=
 # Log level: trace|debug|info|warn|error
 LOG_LEVEL=info
 # Set to "true" to pretty-print logs (dev only).
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -46,6 +46,10 @@ Set at least: `CLAUDE_CODE_OAUTH_TOKEN`, `DISCORD_BOT_TOKEN`,
 default — after startup, message the bot as a super admin and use
 `add_member` / `grant_admin` to onboard people.
 
+Consider also setting `INTERACTION_RETENTION_DAYS` (e.g. `90`) to
+automatically age-purge raw message content per your privacy policy — it's
+disabled by default so existing deployments see no behaviour change.
+
 ## 5. Run migrations
 ```bash
 cd /opt/community-agent
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -105,8 +105,15 @@ A normal user tries to get the agent to moderate, announce, or reveal secrets.
   AI assistant logs interactions (Discord/WhatsApp etiquette + NZ Privacy Act
   2020 expectations).
 - Provide a deletion path: delete rows from `interactions` (and `knowledge`)
-  by `user_id` on request. Consider a retention policy (e.g. purge raw
-  `interactions` older than N months while keeping aggregate `knowledge`).
+  by `user_id` on request (`forget_me` / `purge_user_data`).
+- **Retention policy**: set `INTERACTION_RETENTION_DAYS` to age-purge raw
+  `interactions` (default unset = disabled, no behaviour change on upgrade).
+  A daily in-process timer (`src/index.ts`) deletes rows older than the
+  configured window and logs the count purged. Must be `0` or **at least 7
+  days** (enforced at startup) so a low value can't silently gut memory
+  recall for users still mid-conversation. `knowledge` (curated, durable
+  facts), `admin_audit` (accountability trail), and `sessions` (governed by
+  `SESSION_MAX_TURNS`/`_AGE_HOURS`) are never touched by this purge.
 
 ## Platform-specific notes
 
@@ -198,5 +205,6 @@ number could reach an unrelated person).
 - [ ] Postgres is not exposed to the network.
 - [ ] Bot has minimal Discord permissions.
 - [ ] Community is informed that interactions are logged.
-- [ ] A retention/deletion policy is defined.
+- [ ] A retention/deletion policy is defined (`forget_me`/`purge_user_data`
+      for per-user requests; `INTERACTION_RETENTION_DAYS` for age-based purge).
 - [ ] `journalctl -u community-agent` reviewed for redaction leaks.
--- a/src/config.ts
+++ b/src/config.ts
@@ -59,12 +59,21 @@ const EnvSchema = z.object({
   // Session hygiene: start a fresh Claude session past either cap.
   SESSION_MAX_TURNS: z.coerce.number().int().positive().default(30),
   SESSION_MAX_AGE_HOURS: z.coerce.number().positive().default(24),
+  // Age-based purge of raw `interactions` content. Unset/0 = disabled (no
+  // behaviour change on upgrade). knowledge/admin_audit/sessions are never
+  // touched by this — see storage/repository.ts:purgeOldInteractions.
+  INTERACTION_RETENTION_DAYS: z.coerce.number().int().nonnegative().default(0),
   LOG_LEVEL: z.enum(['trace', 'debug', 'info', 'warn', 'error']).default('info'),
   LOG_PRETTY: z
     .string()
     .optional()
     .transform((v) => v === 'true'),
 });
+
+// Retention must stay well clear of the active-conversation window
+// (SESSION_MAX_AGE_HOURS) so a low value can't silently gut memory recall
+// for users still mid-conversation.
+const MIN_INTERACTION_RETENTION_DAYS = 7;
 
 const EnvSchemaChecked = EnvSchema.refine(
   (e) =>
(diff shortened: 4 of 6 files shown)


swampratnz:
Reviewed. Solid, privacy-positive change:

- **SQL**: `purgeOldInteractions` binds `days` as a real parameter (`$1::text || ' days'`) rather than string-interpolating it into the query — no injection surface, and the value is server-config-derived anyway (not user input).
- **Config guard**: the 7-day floor is enforced via `zod` `.refine()` at startup (fails fast with a readable error) rather than left as an unchecked int — good, and the reasoning (stay clear of `SESSION_MAX_AGE_HOURS`) is sound.
- **Scope discipline**: confirmed `purgeOldInteractions` only touches `interactions` — `knowledge`, `admin_audit`, and `sessions` are untouched, matching the stated intent and existing SECURITY.md residual-risk notes (audit trail retained deliberately).
- **Zero-touch on upgrade**: default `0` means existing deployments get no behavior change, and the timer function itself early-returns `null` when disabled — clean.
- **Timer hygiene**: `setInterval(...).unref()` + cleared in the shutdown handler, consistent with the router's existing sweep-timer pattern.
- CI green.

One non-blocking perf note: `interactions` has no index with `created_at` as a leading column (`schema.sql` only has `(platform, conversation_id, created_at DESC)` and `(platform, user_id, created_at DESC)` composites), so this daily `DELETE FROM interactions WHERE created_at < ...` will do a full table scan rather than an index range scan. Fine at current/expected community scale, but worth a `CREATE INDEX ON interactions (created_at)` if the table grows large enough that this becomes a noticeable lock/IO spike on the daily run — not blocking for this PR.

LGTM — ready for human merge.

---

PR #14, opened 2026-07-03
Add /healthz endpoint and sustained-disconnect super-admin alerting

### LINKED ISSUE #11: Proposal: health endpoint + super-admin alert on sustained platform disconnects
## Problem

The service has no observable signal for "the process is alive but a
platform connection is dead." Concretely:

- `deploy/community-agent.service` has `Restart=always` on **process** exit,
  and `src/index.ts` runs `healthcheck()` (`src/storage/db.ts`) once at
  **startup** — neither covers the steady-state case where the Node process
  keeps running but a platform link has silently dropped.
- `BaileysAdapter` (`src/platforms/whatsapp/baileysAdapter.ts:48-71`) already
  has exponential-backoff reconnect logic and logs each attempt
  (`logger.warn`), but that only surfaces via `journalctl` — nobody is
  paged. A banned/unlinked WhatsApp number (the exact risk `docs/SECURITY.md`
  calls out) would sit in an endless reconnect loop, process healthy,
  service "running," and the community just... stops getting WhatsApp
  replies until someone happens to check logs.
- Discord's gateway reconnects are handled inside `discord.js` with no
  explicit handler in `DiscordAdapter` surfacing sustained failures either.
- Every other "something an operator needs to know about" path in this
  project (privileged actions) already DMs super admins
  (`src/agent/tools.ts`'s `audited()` wrapper, per `docs/SECURITY.md` §"Super-
  admin alerting"). Platform connectivity loss has no equivalent, despite
  being at least as operationally important as an audit event.
- There's also no HTTP surface for external uptime monitoring (a load
  balancer health check, an uptime-robot-style external ping) except the
  WhatsApp Cloud API's webhook port, which only exists in `cloud` mode and
  isn't meant as a general liveness probe.

## Proposed approach

1. **Track adapter connection state** — each `PlatformAdapter` implementation
   exposes a simple `isConnected(): boolean` (Discord: gateway ready +
   `client.ws.status`; WhatsApp Baileys: derived from the existing
   `connection.update` handler at `baileysAdapter.ts:102`; WhatsApp Cloud:
   effectively always "connected" since it's a stateless webhook receiver —
   document that distinction rather than faking a signal).
2. **Sustained-disconnect alert** — a lightweight periodic check (reuse the
   pattern from the proposed retention-purge timer in #9, or a simple
   `setInterval`) that, if an adapter has been disconnected continuously for
   longer than a threshold (e.g. `HEALTH_ALERT_AFTER_MINUTES`, default ~5,
   comfortably longer than Baileys' own backoff ceiling so a normal blip
   doesn't page anyone), sends **one** DM to super admins via the adapter(s)
   that are still up (reuse the existing super-admin-alert send path) and
   logs at `error`. Debounce so it fires once per outage, not once per check
   interval.
3. **`/healthz` HTTP endpoint** — small, unauthenticated, minimal-info
   (`{status: "ok"|"degraded", db: bool, adapters: {discord: bool, whatsapp:
   bool}}`) endpoint for external monitoring. Bind to localhost by default
   (`HEALTH_PORT`, optional — unset disables it, matching the opt-in pattern
   used for other new config in this pipeline); when WhatsApp Cloud API mode
   is active, this can share the existing webhook HTTP listener
   (`WHATSAPP_CLOUD_WEBHOOK_PORT`) instead of opening a second port — worth
   confirming which is simpler during implementation.

## Alternatives considered

- **Do nothing; rely on `journalctl` + `Restart=always`** — status quo;
  rejected because it only catches process crashes, not the "alive but deaf"
  failure mode, which is exactly the WhatsApp-ban scenario the project's own
  docs treat as the top operational risk.
- **systemd watchdog (`sd_notify` + `WatchdogSec=`)** — a real complementary
  option for process-level liveness (catches full hangs), but doesn't solve
  "process fine, one adapter dead" since the main loop keeps ticking; would
  need the same per-adapter tracking regardless. Could be added alongside
  this, not instead of it.
- **Third-party uptime/monitoring service integration (Healthchecks.io,
  UptimeRobot, etc.)** — the `/healthz` endpoint proposed here is exactly
  what those services poll; deliberately not baking in a specific vendor,
  just exposing the primitive.
- **Push metrics to Prometheus/Grafana** — proper observability stack, but a
  large increase in scope (new dependency, new deployment surface) for what
  a single small community bot needs; a health endpoint + DM alert covers
  the actual failure mode without it.

## Security / privacy impact

- `/healthz` returns only boolean connectivity flags — no message content,
  no user data, no internal ids. Still worth binding to localhost by default
  (`HEALTH_PORT` unset = disabled, matching this pipeline's established
  "new surface is opt-in" pattern) so operators explicitly choose to expose
  it, e.g. behind the same reverse proxy already recommended for the Cloud
  API webhook in `docs/DEPLOYMENT.md`.
- No new privileged tool, no RBAC surface — the alert reuses the existing
  super-admin DM channel and the existing `audited()`-style send path, not a
  new one.
- No user-controllable input on the health check path — nothing to inject.

## Rough scope

- `src/platforms/types.ts` — add `isConnected(): boolean` to
  `PlatformAdapter`.
- `src/platforms/discord/adapter.ts`, `src/platforms/whatsapp/baileysAdapter.ts`,
  `src/platforms/whatsapp/cloudAdapter.ts` — implement it (cloud: constant
  `true` with a comment explaining why).
- `src/router.ts` or `src/index.ts` — periodic check + debounced super-admin
  alert; `src/config.ts` — `HEALTH_ALERT_AFTER_MINUTES`, `HEALTH_PORT`
  (both optional, sane defaults / disabled-by-default for the port).
- Small HTTP handler for `/healthz` (native `http`, no new dependency needed
  for one route).
- `docs/DEPLOYMENT.md` — document the endpoint and alert threshold;
  `docs/SECURITY.md` — note the new (opt-in, localhost-default) HTTP surface
  alongside the existing Cloud API webhook note.
- Tests: debounce logic (alert fires once per sustained outage, not
  repeatedly), `/healthz` reflects adapter state, disabled-by-default when
  `HEALTH_PORT` unset.

## Acceptance criteria

- [ ] A WhatsApp Baileys disconnect lasting longer than the threshold
      produces exactly one super-admin DM alert (via Discord, if that
      adapter is still up) and an `error`-level log line.
- [ ] Recovery (reconnect) does not itself spam an additional alert.
- [ ] `/healthz` (when `HEALTH_PORT` is set) reports accurate `db` and
      per-platform connectivity; unset by default, no new listening port on
      unchanged deployments.
- [ ] No message content or user identifiers appear in the health response
      or the alert DM beyond "which platform is down."

swampratnz:
Closes #11

## Summary

The service had no steady-state signal for "process alive, one platform connection silently dead." `Restart=always` only catches process crashes, and `healthcheck()` only runs once at startup — neither catches a WhatsApp number stuck in Baileys' reconnect loop (the exact ban scenario `docs/SECURITY.md` calls out as the top operational risk) with the process otherwise healthy and nobody paged.

- **`src/platforms/types.ts`** — added `isConnected(): boolean` to `PlatformAdapter`.
- **Adapters**:
  - `DiscordAdapter` — tracks a `connected` flag via `Events.ClientReady`/`ShardDisconnect`/`ShardResume` (discord.js resumes gateway sessions internally, but a shard disconnect means we're not receiving messages, which is what this cares about).
  - `BaileysAdapter` — tracks via the existing `connection.update` handler's `'open'`/`'close'` branches (no new event wiring needed).
  - `WhatsAppCloudAdapter` — always `true` while its HTTP listener is up; it's a stateless webhook receiver with no persistent connection to track the way the other two have, documented explicitly rather than faking a signal.
- **`src/healthState.ts`** (new, pure, no config/HTTP/adapter imports — mirrors the `wire.ts`/`cloudWire.ts` convention so it's directly unit-testable):
  - `stepDisconnectTracker` — the debounce state machine: an alert fires once per sustained outage past the threshold, not once per check tick, and reconnecting resets silently (no "it's back!" spam).
  - `buildHealthzPayload` — `{status: "ok"|"degraded", db, adapters}` shape.
- **`src/health.ts`** (new, I/O wrapper around the above):
  - `startDisconnectAlerts` — a 30s periodic check across all registered adapters; past `HEALTH_ALERT_AFTER_MINUTES` (default 5) of continuous disconnection, DMs configured super admins via whichever adapter(s) are still connected (reusing `sendDirectMessage`, the same send path `tools.ts`'s existing alerting already uses) and logs at `error`.
  - `startHealthServer` — opt-in (`HEALTH_PORT` unset = no listening port at all) unauthenticated `GET /healthz` returning boolean connectivity flags only, no message content or user ids.
- **`src/index.ts`** — wires both in after adapters start; the health server and alert timer are cleanly torn down on graceful shutdown.
- **Docs** — `docs/ARCHITECTURE.md` (new "Health & monitoring" section + components table), `docs/SECURITY.md` (new opt-in-HTTP-surface note alongside the existing Cloud API webhook one), `docs/DEPLOYMENT.md`, `.env.example`.

## Security impact

- `/healthz` is opt-in (`HEALTH_PORT` unset = disabled by default, matching this pipeline's established "new surface is opt-in" pattern) and returns only boolean connectivity flags — no message content, no user identifiers, no internal ids.
- No new RBAC surface or privileged tool — the disconnect alert reuses the existing `sendDirectMessage` adapter method, the same primitive `tools.ts`'s `notifySuperAdmins` already uses for privileged-action alerts.
- No user-controllable input anywhere on this path — nothing to inject.

## Verification

- `npm test` — 62/62 passing (53 existing + 9 new in `tests/healthState.test.ts`), no regressions. Covers: no alert while connected, no alert under the debounce threshold, exactly one alert once the threshold is crossed, no repeat alert while still down, reconnecting resets the tracker without alerting, a fresh outage after recovery can alert again, and the `/healthz` payload shape (ok/degraded transitions, and that it carries no message-content/user-id fields).
- `npm run build` — clean.
- **Exercised against a local Postgres 16 + pgvector** (per CLAUDE.md) and real timers via the built `dist/`, with fake `PlatformAdapter`s standing in for live sockets:
  - `/healthz`: all-connected + real DB → `200 {status: "ok", db: true, adapters: {...}}`; flipping one adapter's `isConnected()` to `false` → `503 {status: "degraded", ...}`; an unknown path → `404`.
  - Disconnect alerting: with no super admins configured, the loop runs across real 30s check ticks without throwing and sends nothing. With `SUPER_ADMIN_DISCORD_IDS` configured and WhatsApp down from the start, waited across real ticks and confirmed exactly one alert DM (log line `Platform sustained disconnect` + `sendDirectMessage` calls) fires once the debounce threshold is crossed, and does **not** repeat on the following tick while still down.
- Did not exercise `isConnected()` against a live Discord gateway or WhatsApp socket (would require a real bot token / linked number) — the event wiring mirrors the existing, already-relied-upon `connection.update`/`ShardDisconnect` handlers already present in each adapter for reconnect logic and logging.
- No automated DB-integration test added beyond the pure `healthState.ts` suite — CI has no Postgres service; matches the convention established across the three prior PRs in this pipeline (#8, #10, #12).

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -69,6 +69,13 @@ SESSION_MAX_AGE_HOURS=24
 # Age-based purge of raw `interactions` (privacy/retention). Unset/0 = disabled.
 # Must be 0 or >= 7 if set. knowledge/admin_audit/sessions are never touched.
 INTERACTION_RETENTION_DAYS=
+# A platform disconnected continuously past this many minutes triggers one
+# debounced super-admin DM alert (via whichever adapter is still up).
+HEALTH_ALERT_AFTER_MINUTES=5
+# Unauthenticated GET /healthz -> {status, db, adapters}. Unset = disabled
+# (no listening port). If set, bind to localhost and put a reverse proxy in
+# front — same guidance as the WhatsApp Cloud webhook port.
+HEALTH_PORT=
 # Log level: trace|debug|info|warn|error
 LOG_LEVEL=info
 # Set to "true" to pretty-print logs (dev only).
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -52,6 +52,7 @@ Claude-powered agent, with a Postgres-backed memory for learning.
 | `src/agent/auth.ts` | Forces Claude **subscription** auth via `CLAUDE_CODE_OAUTH_TOKEN`. |
 | `src/storage/*` | Postgres pool, schema, migrations, embeddings, repository. |
 | `src/router.ts` | Orchestrates inbound → agent → outbound and persistence. |
+| `src/health.ts` / `src/healthState.ts` | `/healthz` endpoint + sustained-disconnect super-admin alerting; `healthState.ts` holds the pure, tested debounce/payload logic. |
 
 ## Memory & "learning"
 
@@ -153,6 +154,35 @@ a few seconds of overhead per answered message, growing with session length.
 If this becomes a problem: cap session length (start fresh after N turns), or
 move to the SDK's streaming-input mode with a persistent process per busy
 conversation.
+
+## Health & monitoring
+
+`Restart=always` (`deploy/community-agent.service`) and the startup
+`healthcheck()` only catch the process crashing or the DB being unreachable
+at boot — neither catches "process alive, one platform connection silently
+dead" (e.g. a banned WhatsApp number stuck in Baileys' reconnect loop).
+`src/health.ts` covers that steady-state gap:
+
+- **Sustained-disconnect alerting** (always on, no config to disable) — a
+  30s periodic check across every registered adapter's `isConnected()`. Past
+  `HEALTH_ALERT_AFTER_MINUTES` (default 5) of continuous disconnection, it
+  DMs configured super admins via whichever adapter(s) are still up and logs
+  at `error`. Debounced: one alert per outage, not one per check tick;
+  reconnecting clears the state silently (no "it's back!" spam).
+- **`/healthz`** (opt-in via `HEALTH_PORT`) — unauthenticated `GET` returning
+  `{status: "ok"|"degraded", db: boolean, adapters: {discord: boolean,
+  whatsapp: boolean}}`. No message content or user ids in the response.
+  Intended for an external uptime monitor; bind to localhost and put a
+  reverse proxy in front if exposing it, same guidance as the Cloud API
+  webhook port.
+- `WhatsAppCloudAdapter.isConnected()` reflects whether its local HTTP
+  listener is up, not whether Meta can currently reach it — it's a
+  stateless webhook receiver with no persistent connection to track the way
+  Baileys/Discord have.
+
+The debounce/payload logic lives in `src/healthState.ts`, deliberately free
+of config/HTTP/adapter imports so it's unit-tested directly (`src/health.ts`
+is the thin I/O wrapper around it).
 
 ## Switching WhatsApp providers
 
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -50,6 +50,12 @@ Consider also setting `INTERACTION_RETENTION_DAYS` (e.g. `90`) to
 automatically age-purge raw message content per your privacy policy — it's
 disabled by default so existing deployments see no behaviour change.
 
+Sustained platform-disconnect alerting to super admins is always on
+(`HEALTH_ALERT_AFTER_MINUTES`, default 5 minutes). Optionally set
+`HEALTH_PORT` to expose an unauthenticated `GET /healthz` for an external
+uptime monitor — bind it to localhost and reverse-proxy it if you expose it
+publicly, same as the WhatsApp Cloud API webhook.
+
 ## 5. Run migrations
 ```bash
 cd /opt/community-agent
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -139,6 +139,16 @@ parsed before that check passes. `WHATSAPP_CLOUD_ACCESS_TOKEN` and
 delivery metadata for Cloud API traffic are additionally retained by Meta
 per their own terms, on top of this project's own storage.
 
+### `/healthz` endpoint
+Opt-in (`HEALTH_PORT` unset = no listening port at all — matches this
+pipeline's "new surface is opt-in" pattern). Unauthenticated by design, but
+the response is boolean connectivity flags only (`{status, db, adapters}`) —
+no message content, no user identifiers, no internal ids. Bind to localhost
+and put a reverse proxy in front if exposing it externally, same guidance as
+the Cloud API webhook port above. The sustained-disconnect super-admin DM
+alert reuses the existing adapter `sendDirectMessage` path — no new
+privileged tool, no RBAC surface.
+
 ### Discord
 - Enable only the gateway intents the bot needs (Guilds, GuildMessages,
   MessageContent, GuildMembers, DirectMessages). **Both `MessageContent` and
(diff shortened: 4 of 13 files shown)


swampratnz:
Reviewed. Good addition, closes a real operational gap (silent WhatsApp reconnect-loop with no page).

- **No RBAC/new privileged surface**: the disconnect alert reuses `sendDirectMessage` via `superAdminIds(platform)` — same primitive as existing privileged-action alerting, no new tool, nothing for `assertAtLeast` to gate.
- **`/healthz` payload is minimal**: `{status, db, adapters}` — booleans only, confirmed by the explicit "no message content or user identifiers in the shape" test. No injection surface (no request body/params consumed beyond method+path).
- **Debounce logic is correctly isolated and well-tested**: `healthState.ts` is pure, and the test suite covers the cases that actually matter for an alerting system — exactly one alert per sustained outage, no repeat while still down, silent reset on reconnect, and a fresh outage after recovery can alert again. That's the right set of edge cases for a debounce state machine.
- **Adapter wiring is consistent**: Discord tracks `ShardDisconnect`/`ShardResume`, Baileys reuses the existing `connection.update` open/close branches (no new event surface), and the Cloud adapter's `isConnected()` is honestly scoped to "local listener up," not "Meta can reach us" — documented rather than faked.
- CI green (62/62).

One non-blocking note, same pattern as the existing WhatsApp Cloud webhook: docs say `/healthz` should "bind to localhost," but `startHealthServer`'s `server.listen(port, callback)` doesn't pass a host, so it listens on all interfaces (`0.0.0.0`) by default — same as `cloudAdapter.ts`'s webhook listener. For the Cloud webhook that's necessary (Meta needs to reach it publicly), but `/healthz` is meant to be *internal-only* per the docs, so relying purely on operator firewalling to enforce that is a bit of a gap between the stated guidance and what the code guarantees. Worth considering `server.listen(port, '127.0.0.1', callback)` for defense-in-depth so an operator who forgets the reverse-proxy/firewall step doesn't end up with an unauthenticated endpoint reachable from the internet — low severity since the payload itself is just connectivity booleans, but cheap to fix.

LGTM — ready for human merge.

---

PR #15, opened 2026-07-03
Add Postgres+pgvector CI service and repository-level integration tests

### LINKED ISSUE #13: Proposal: Postgres+pgvector service in CI and repository-level integration tests
## Problem

`.github/workflows/ci.yml` runs `npm run typecheck && npm test && npm run
build` on every push/PR — no Postgres service, no `DATABASE_URL`. `npm test`
(`tsx --test tests/*.test.ts`) covers `agentOptions`, `outbound`,
`pendingActions`, `rbac`, `systemPrompt`, and the WhatsApp wire helpers — real
and valuable, but every one of them is pure logic with no database. Nothing
in `tests/` imports or exercises `src/storage/repository.ts` or
`src/storage/schema.sql` at all.

That file is now the single most load-bearing piece of the codebase and it's
grown fast and entirely untested by CI:

- `save_knowledge`/`list_knowledge`/`update_knowledge`/`delete_knowledge` —
  admin-facing SQL, added in #5.
- `purgeOldInteractions` — an actual `DELETE` running on a schedule in
  production, added in #9. A wrong `WHERE` clause here silently deletes the
  wrong rows, in prod, with no test catching it first.
- `recordAccessRequest`/`listAccessRequests` — added in #7.
- `purgeUserData`/cascading deletes across `interactions`+`knowledge`, the
  embedding-backed `remember_search`/`knowledge_search` (pgvector HNSW
  cosine queries), `admin_audit`, `sessions`, `usage_stats` aggregation.

CLAUDE.md already states the standard: "DB-touching changes: exercise against
a local Postgres 16 + pgvector" — but that's currently a manual step a human
or an agent has to remember to do locally; it is not enforced anywhere, and
nothing stops a PR with a broken `DELETE`/`UPDATE`/scoping query from merging
green. Given the project's own security posture leans hard on "admin data
access is scoped in SQL" (CLAUDE.md) and multiple purge/delete paths now
exist, an untested query layer is a real correctness *and* security gap: a
subtly wrong scope filter (e.g. an admin's cross-conversation query missing
its scoping clause) would currently ship with a fully green CI run.

## Proposed approach

1. **Add a Postgres+pgvector service container to `.github/workflows/ci.yml`**
   — `pgvector/pgvector:pg16` (matches `docs/DEPLOYMENT.md`'s "PostgreSQL +
   pgvector (0.8.4+)"), with a health check, standard `postgres`/`postgres`
   creds, exposed on `5432`. Run `npm run migrate` (already exists,
   `src/storage/migrate.ts`) against it before tests.
2. **New `tests/repository.test.ts`** (or a `tests/db/` subfolder, matching
   existing flat structure unless it gets large) covering the highest-risk
   paths concretely:
   - `purgeOldInteractions`: rows older than cutoff removed, newer rows and
     `knowledge`/`admin_audit`/`sessions` untouched (the exact acceptance
     criteria #9 already specified but that PR's tests should be checked —
     if they used a real DB harness already, this proposal formalises/wires
     it into CI rather than re-inventing it).
   - `purgeUserData`/`forget_me` cascade: deletes the right rows, leaves
     other users' data and `admin_audit` intact.
   - `knowledge` CRUD: insert → search finds it, update re-embeds and bumps
     `updated_at`, delete removes it from subsequent search results.
   - Admin cross-conversation scoping queries: an admin's query only returns
     rows for conversations they're actually scoped to (this is the one
     CLAUDE.md calls out by name as a security invariant — it deserves an
     actual assertion against real SQL, not just code review).
3. **Local dev parity** — `npm run test:db` or extend `npm test` to skip
   DB-dependent tests gracefully when `DATABASE_URL` is unset (so contributors
   without local Postgres aren't blocked), matching how `tsx --test` already
   partitions by file.
4. Update CLAUDE.md's "Build / test / verify" section once this lands: the
   manual "exercise against a local Postgres 16 + pgvector" instruction
   becomes "CI does this for you; do it locally too for the tight loop."

## Alternatives considered

- **Mock the `pg` `Pool`/query results instead of a real DB** — faster, no
  service container, but exactly the kind of test that would have passed a
  wrong `WHERE` clause with a mismatched cutoff or a missing scope filter,
  since the mock encodes the same assumption the code makes. Rejected for
  the highest-risk queries (purge, admin scoping) — that's precisely where a
  real DB is worth the CI minutes. Lighter unit tests with mocks remain fine
  for anything that doesn't touch SQL correctness.
- **Leave it manual, just document it better** — status quo; rejected, this
  is exactly the gap CLAUDE.md already flags without enforcing, and the
  project has shipped four PRs' worth of new repository functions since
  launch with zero CI coverage on any of them.
- **Full testcontainers-style ephemeral DB per test file** — more isolation,
  but adds a dependency and complexity; a single CI-wide service container
  (start once, `TRUNCATE` or transaction-rollback between tests) is simpler
  and matches the project's minimal-dependency philosophy (no test
  framework beyond `node:test` today).

## Security / privacy impact

- Directly **improves** the security posture: the admin cross-conversation
  scoping invariant (CLAUDE.md: "Admin data access is scoped in SQL to
  conversations the admin is in") currently has no automated regression
  test at the SQL layer — a future change could silently widen that scope
  and CI would stay green. This proposal closes that.
- No new production surface — CI-only infrastructure change plus test files.
  No secrets involved (CI Postgres uses throwaway local creds, never touches
  real data).
- Test data must be synthetic — never real community member data, obviously
  (goes without saying, but worth stating for the acceptance criteria: no
  production DB dump ever becomes a fixture).

## Rough scope

- `.github/workflows/ci.yml` — add `services: postgres:` block +
  `DATABASE_URL` env for the test step + a migrate step before `npm test`.
- `tests/repository.test.ts` (+ maybe a small `tests/helpers/db.ts` for
  setup/teardown/truncate-between-tests).
- `package.json` — optional `test:db` script or a `DATABASE_URL`-gated skip
  inside the existing `test` script, whichever keeps `npm test` a single
  command that works both locally-without-DB and in CI.
- `CLAUDE.md` — update the build/test/verify section to reflect CI now
  covering this.

## Acceptance criteria

- [ ] CI spins up `pgvector/pgvector:pg16`, runs migrations, and executes
      repository-level tests against it on every PR.
- [ ] `purgeOldInteractions` has a real-DB test proving it deletes only rows
      past the cutoff and leaves `knowledge`/`admin_audit`/`sessions` alone.
- [ ] `purgeUserData`/`forget_me` cascade has a real-DB test proving scoping
      (only the target user's data, nothing cross-user).
- [ ] At least one test asserts the admin cross-conversation scoping filter
      actually excludes conversations the admin isn't in.
- [ ] `npm test` still passes locally for a contributor with no
      `DATABASE_URL` set (DB tests skip cleanly, don't fail the run).
- [ ] `npm run typecheck`, `npm test`, `npm run build` all green in CI with
      the new service container.

swampratnz:
Closes #13

## Summary

`src/storage/repository.ts` had grown into the most load-bearing, security-relevant file in the project — knowledge CRUD, the `purgeOldInteractions` DELETE that runs on a schedule in production, `access_requests`, and the admin cross-conversation scoping filter CLAUDE.md names by name as a security invariant — with **zero** CI coverage. Every existing test in `tests/` is pure logic with no database.

- **`.github/workflows/ci.yml`** — added a `pgvector/pgvector:pg16` service container (matches `docs/DEPLOYMENT.md`'s recommended version), health-checked via `pg_isready`, exposed on `5432`. The job now sets dummy `CLAUDE_CODE_OAUTH_TOKEN`/`DISCORD_BOT_TOKEN`/`DISCORD_GUILD_ID` (required by `config.ts`'s zod schema regardless of what a given test needs) plus a real `DATABASE_URL` pointing at the service container, runs `npm run migrate` before `npm test`.
- **`tests/repository.test.ts`** (new) covers the four highest-risk paths the proposal called out:
  - `purgeOldInteractions` deletes only rows older than the cutoff; `knowledge` and `admin_audit` are never touched regardless of age.
  - `purgeUserData` cascade only affects the target user's `interactions` + knowledge sourced from them; another user's data and the audit trail are untouched.
  - `knowledge` CRUD: insert → `searchKnowledge` finds it; `updateKnowledge` re-embeds (verified via search on the new content) and bumps `updated_at` while preserving unspecified fields; `deleteKnowledge` removes it and returns `false` (not an error) on a second call.
  - **`SECURITY:`** the admin cross-conversation scoping filter (`userMessages`'s `conversationIds` param, backing the `user_history` tool) actually excludes a conversation outside the given scope — this is the invariant CLAUDE.md calls out by name that previously had no automated regression test at the SQL layer.
- **Local dev parity**: DB-touching tests check `Boolean(process.env.DATABASE_URL)` and pass `{ skip }` to `test()` when it's unset, so `npm test` still passes cleanly (skipped, not failed) for a contributor without local Postgres — matching the existing `tests/agentOptions.test.ts` convention for stubbing `config.ts`'s required env vars before import.
- **`CLAUDE.md`** — updated the "Build / test / verify" section: DB-touching changes are now enforced by CI, not just a manual reminder, with the local-dev-parity instructions.

One deliberate safety choice: `purgeOldInteractions` is an **un-scoped, whole-table** delete by design (that's its actual production behavior). To make the test safe to run against a populated local dev Postgres (not just the throwaway CI container), the test backdates its fixtures to ~100 years old (`36,525`/`36,526` days) rather than a realistic retention window like 30 days, and asserts the exact boundary (`created_at` just under vs. just over the cutoff). Since no real interaction can predate the schema itself, this makes the test's destructive `DELETE` provably incapable of touching legitimate recent data, while still exercising the exact SQL comparison the function performs.

## Security impact

- **Improves** the security posture: closes the gap CLAUDE.md already names ("Admin data access is scoped in SQL to conversations the admin is in") but had no automated test proving — a future change could silently widen that scope and CI would have stayed green.
- No new production surface — CI-only infrastructure (a service container) plus test files. The CI Postgres uses throwaway, non-secret local credentials (`postgres`/`postgres`) that never touch real data.
- Test fixtures are synthetic, tagged with a per-run unique id, and fully cleaned up at the end of each test (verified manually — see below) — no risk of leftover data accumulating in a real Postgres a contributor points `DATABASE_URL` at locally.

## Verification

- `npm test` **without** `DATABASE_URL` set — 62 pass, 0 fail, **4 skipped** (the new DB tests), confirming the local-dev-parity acceptance criterion (a contributor without local Postgres isn't blocked).
- `npm test` **with** `DATABASE_URL` set against a real local Postgres 16 + pgvector (per CLAUDE.md) — **66/66 pass, 0 skipped**, including all 4 new repository tests running for real against live SQL.
- `npm run build` — clean.
- Manually confirmed via `psql` that after the full test run, `interactions`, `knowledge`, and `admin_audit` all have **0 rows** — every fixture the new tests create is cleaned up, leaving no residue in the test database.
- Manually sanity-checked `.github/workflows/ci.yml`'s YAML is well-formed (`yaml.safe_load`).
- Did not run this against the actual GitHub Actions runner (would require pushing and watching the Actions tab) — the local verification above exercises the identical `npm run migrate && npm test` sequence the workflow runs, against the same `pgvector/pgvector:pg16`-equivalent local setup (Postgres 16 + the `vector` extension installed), so I'm confident it'll behave the same in CI; will confirm once this PR's own CI run completes.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,28 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: pgvector/pgvector:pg16
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: community_agent_test
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    env:
+      # Dummy values so config.ts's zod schema is satisfied — nothing in
+      # this job talks to Claude or Discord, only DATABASE_URL is real.
+      CLAUDE_CODE_OAUTH_TOKEN: ci-dummy-token
+      DISCORD_BOT_TOKEN: ci-dummy-token
+      DISCORD_GUILD_ID: ci-dummy-guild
+      WHATSAPP_PROVIDER: disabled
+      DATABASE_URL: postgres://postgres:postgres@localhost:5432/community_agent_test
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -16,5 +38,6 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npm run typecheck
+      - run: npm run migrate
       - run: npm test
       - run: npm run build
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,7 +16,13 @@ with `README.md`, then `docs/ARCHITECTURE.md` and `docs/SECURITY.md`.
   here (tool gating, confirm flow, secret redaction, WhatsApp wire helpers) —
   when you touch those areas, extend the tests.
 - `npm run build` — tsc + copies `schema.sql` into `dist/`.
-- DB-touching changes: exercise against a local Postgres 16 + pgvector.
+- DB-touching changes: CI runs `tests/repository.test.ts` against a real
+  `pgvector/pgvector:pg16` service container (see `.github/workflows/ci.yml`),
+  so this is enforced, not just a manual reminder. Do it locally too for the
+  tight loop: run `npm run migrate` against a local Postgres 16 + pgvector
+  with `DATABASE_URL` set, then `npm test` — DB-touching tests skip cleanly
+  (not fail) when `DATABASE_URL` is unset, so a contributor without local
+  Postgres isn't blocked.
 - Run all three (typecheck, test, build) green before opening/updating a PR.
 
 ## Security posture (do not regress)
--- a/tests/repository.test.ts
+++ b/tests/repository.test.ts
@@ -1,0 +1,235 @@
+import { test, after } from 'node:test';
+import assert from 'node:assert/strict';
+
+// config.ts validates env at import time — provide a dummy environment
+// before importing anything that (transitively) loads it, matching the
+// convention in tests/agentOptions.test.ts.
+const hasDb = Boolean(process.env.DATABASE_URL);
+
+process.env.CLAUDE_CODE_OAUTH_TOKEN ??= 'test-token';
+process.env.DISCORD_BOT_TOKEN ??= 'test-token';
+process.env.DISCORD_GUILD_ID ??= '1';
+process.env.DATABASE_URL ??= 'postgres://test:test@localhost:5432/test';
+process.env.WHATSAPP_PROVIDER ??= 'disabled';
+
+const skip = hasDb ? false : 'DATABASE_URL not set — skipping DB-integration tests (CLAUDE.md: exercise against a local Postgres 16 + pgvector)';
+
+const { pool, closeDb } = await import('../src/storage/db.js');
+const {
+  recordInteraction,
+  userMessages,
+  purgeOldInteractions,
+  purgeUserData,
+  saveKnowledge,
+  searchKnowledge,
+  updateKnowledge,
+  deleteKnowledge,
+  recordAdminAction,
+} = await import('../src/storage/repository.js');
+
+// Unique per test-run tag so fixtures never collide across runs and can be
+// cleaned up precisely, whether this runs against a throwaway CI Postgres or
+// a developer's real local instance.
+const RUN = `t${Date.now()}${Math.floor(Math.random() * 1e6)}`;
+
+after(async () => {
+  await closeDb();
+});
+
+test(
+  'repository: purgeOldInteractions deletes only rows past the cutoff, never knowledge/admin_audit',
+  { skip },
+  async () => {
+    const userId = `${RUN}-purge`;
+    const conversationId = `${RUN}-c-purge`;
+
+    // Use an extreme age (~100 years) rather than a realistic retention
+    // window: this makes the boundary assertions correct regardless of what
+    // else lives in the table (real interactions can't predate the schema),
+    // so this test is safe to run against a populated local dev DB too.
+    const HUNDRED_YEARS_DAYS = 36_525;
+
+    await recordInteraction({
+      platform: 'discord',
+      conversationId,
+      userId,
+      role: 'member',
+      direction: 'inbound',
+      content: 'recent — must survive',
+    });
+    await pool.query(
+      `INSERT INTO interactions (platform, conversation_id, user_id, role, direction, content, created_at)
+       VALUES ($1,$2,$3,$4,$5,$6, now() - interval '${HUNDRED_YEARS_DAYS - 1} days')`,
+      ['discord', conversationId, userId, 'member', 'inbound', 'just under the cutoff — must survive'],
+    );
+    await pool.query(
+      `INSERT INTO interactions (platform, conversation_id, user_id, role, direction, content, created_at)
+       VALUES ($1,$2,$3,$4,$5,$6, now() - interval '${HUNDRED_YEARS_DAYS + 1} days')`,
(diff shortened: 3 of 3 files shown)


swampratnz:
Reviewed. This is exactly the kind of gap worth closing — `repository.ts` is the most security-load-bearing file in the project and had zero DB-backed CI coverage.

- **CI service container**: dummy `postgres`/`postgres` creds scoped to an ephemeral GitHub Actions container, never touching real data — appropriate for throwaway CI infra. Dummy `CLAUDE_CODE_OAUTH_TOKEN`/`DISCORD_BOT_TOKEN` are obviously fake placeholders, not real secrets.
- **The `SECURITY:`-tagged test is the standout**: it directly exercises the admin cross-conversation scoping invariant CLAUDE.md calls out by name (`userMessages`'s `conversationIds` param → `AND conversation_id = ANY($n)`, parameterized correctly) — confirmed the test's assertions line up with the actual implementation in `repository.ts:249`. This closes a real regression-detection gap: previously a change that silently widened that scope would have stayed CI-green.
- **`purgeOldInteractions`/`purgeUserData` tests**: correctly assert `knowledge`/`admin_audit` are untouched, and the 100-year backdating trick is a smart way to make the test safe against a populated local dev DB without weakening what it actually proves (exact boundary comparison on the SQL predicate).
- **Local-dev parity**: `{ skip }` when `DATABASE_URL` is unset matches the existing `tests/agentOptions.test.ts` convention — confirmed via the diff that this doesn't block contributors without local Postgres.
- Fixture cleanup at the end of each test, tagged with a per-run unique id — no residue risk.
- CI green.

No issues found. LGTM — ready for human merge.

---

PR #122, opened 2026-07-04
Auto-resolve build-worker PRs that conflict with main

swampratnz:
## Summary

When `main` advances, an open build-worker PR can flip from mergeable to CONFLICTING. GitHub emits **no webhook** for that transition (and `mergeable` is computed asynchronously, so it isn't even known at push time), so today a human has to do the merge by hand. This adds `pipeline-pr-conflict.yml`: on every push to `main` it discovers open same-repo bot PRs that have gone CONFLICTING and has Claude merge `main` into each branch, resolve the conflicts, re-run the full gate, and push — completing the "no human in the routine loop" story alongside the build-retry (#119) and autofix loops.

- **`discover` job** — lists open PRs, keeps same-repo (`!isCrossRepository`) bot-authored PRs that are **not** already `needs-human`, polls `gh pr view --json mergeable` until it settles (it reads `UNKNOWN` right after a push), and emits a matrix of the `CONFLICTING` ones.
- **`resolve` job** — one entry per conflicting PR, run serially (`max-parallel: 1`) against a pgvector Postgres service. Deterministic migrate first, then a `claude-code-action` step that merges `origin/main`, resolves conflicts (preserving both the PR's intent and what's now on `main`), runs the full gate, and pushes. A verify-pushed-else-escalate step is the source of truth.

Docs updated: `CLAUDE.md` and `docs/PIPELINE.md` describe the new loop as the second push-exception (after autofix) to the "only the build loop writes code" rule.

## Security / privacy impact

No change to the bot's runtime auth, tool gating, outbound filtering, or data access — this is CI-only. It does grant a new automated **push** path, so it deliberately mirrors the proven `pipeline-pr-autofix.yml` guardrails:

- Same-repo, bot-authored PRs only — never touches human or fork branches.
- The agent's `gh` is **read-only**; its only push tool is the **exact** `Bash(git push origin HEAD)` (no `:*` wildcard), so it can push only the PR's own branch, fast-forward — never `main`, never `--force`.
- `GH_TOKEN` is set per-step on the deterministic shell steps only, never on the agent step (which reads untrusted PR content), so the agent can't exfiltrate a `GITHUB_TOKEN`.
- Cannot push `.github/workflows/*` (App token lacks the `workflows` scope) — the prompt tells it to escalate instead.
- **One attempt per conflict**: a failed resolution labels `needs-human`, and the `discover` filter skips `needs-human` PRs, so it never thrashes on repeated pushes to `main`.
- Never opens or merges PRs — a human still merges.

Branch protection on `main` (blocking direct/force pushes) remains the recommended repo-setting backstop regardless of token scope.

## How verified

- `python3 -c "import yaml"` parse of the new workflow: top-level keys, both jobs (`discover`, `resolve`), `discover` outputs (`matrix`, `has_conflicts`), and the `resolve` gate (`if: needs.discover.outputs.has_conflicts == 'true'`) all resolve as intended.
- Guardrails and step structure verified against `pipeline-pr-autofix.yml` (identical `allowedTools` push discipline, step-scoped `GH_TOKEN`, deterministic migrate, verify-else-escalate).
- No app/source code changed, so `typecheck`/`test`/`build` are unaffected; the workflow's behaviour validates on the first real conflict.

Note: the bot cannot push `.github/workflows/*`, so this PR needs a human to merge it to go live.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DB9c2BDLYYdRGpghMGHBYo)_

### DIFF
--- a/.github/workflows/pipeline-pr-conflict.yml
+++ b/.github/workflows/pipeline-pr-conflict.yml
@@ -1,0 +1,254 @@
+name: Pipeline PR conflict resolver
+
+# Event-driven MERGE-CONFLICT loop: when `main` advances, an open build-worker
+# PR can flip from mergeable to CONFLICTING. GitHub emits NO webhook for that
+# transition (and `mergeable` is computed asynchronously, so it isn't even known
+# at push time), so we trigger on every push to `main`, discover which same-repo
+# bot PRs have become conflicting, and have Claude merge `main` into each branch,
+# resolve the conflicts, re-run the full gate, and push — so a conflicting PR
+# gets un-stuck without a human doing the merge by hand.
+#
+# Guardrails (mirrors pipeline-pr-autofix.yml — this spends the shared Max pool
+# and pushes code, so they matter):
+#   * Same-repo, bot-authored PRs only. workflow-triggered secrets + a push
+#     tool must never run against human- or fork-controlled branches, so the
+#     discover filter hard-requires author.is_bot && !isCrossRepository.
+#   * PRs already labelled `needs-human` are SKIPPED — that label is the stop.
+#     A failed resolution labels `needs-human`, so each conflict gets exactly
+#     ONE attempt and the loop will not thrash on every subsequent push to main
+#     (which would otherwise drain the weekly token pool re-fighting the same
+#     unresolvable conflict).
+#   * The agent's `gh` is READ-only and its ONLY push tool is the EXACT command
+#     `git push origin HEAD` (no `:*` wildcard) — it can push only the PR's own
+#     branch, fast-forward, never `main` and never `--force` (same reasoning as
+#     the autofix loop).
+#   * GH_TOKEN is set per-step on the deterministic shell steps only, never on
+#     the agent step (which reads untrusted PR/issue content), so the agent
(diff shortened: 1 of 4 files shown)


swampratnz:
Both points addressed in df8182d:

- **Critical (checkout → push-main bypass):** Confirmed and fixed. `git push origin HEAD` pushes to a remote branch named after the *current local* branch, so `Bash(git checkout:*)` would have let the agent `git checkout -B main origin/main` and turn the one permitted push into `git push origin main`. Dropped `git checkout` from `allowedTools` (`branch`/`switch` were never granted); conflict resolution now uses Edit/Write + `git restore --ours/--theirs` + `git add`, so the local branch stays pinned to the PR branch the job checked out. The tool-surface comment now documents this mechanism explicitly rather than overclaiming.

- **Minor (scope):** Tightened the discover filter to require the PR body to reference `Closes #<issue>` — the build worker's contract, and the same signal `pipeline-build.yml`'s verify step keys on — so an unrelated same-repo bot PR (e.g. a Dependabot bump) no longer qualifies for an automated main-merge + push. This matches the documented "build-worker PRs" scope.

The `main`-branch-protection backstop remains recommended regardless, but the tool-level guarantee now holds as claimed again.

---

Issue #90, opened 2026-07-04
Proposal: proactively alert admins when a member files report_content, instead of relying on a manual list_reports pull


swampratnz:
## Problem

Evidence, from reading the shipped feature directly (`src/agent/tools.ts`, `report_content`/`list_reports`/`resolve_report`, landed via #70):

`report_content` (member tier) writes a row to `content_reports` via `createContentReport()` and returns a confirmation to the reporter — full stop. There is **no notification to any admin, anywhere, ever** when a report is filed. Every *other* proactive-alert mechanism in the codebase treats "something happened that a human should know about now" as worth a push, not a pull:

- Every privileged admin action alerts super admins immediately (`notifySuperAdmins`, called from `audited()` in `tools.ts:158-165`).
- A sustained platform disconnect DMs super admins (`src/health.ts`, `startDisconnectAlerts`).
- Usage nearing the Max-pool threshold DMs super admins (`src/usageAlert.ts`).

`report_content` — arguably the most time-sensitive of all of these, since it exists specifically for "a member is being harassed/spammed right now" (per `docs/ARCHITECTURE.md`'s "Abuse reporting" section) — is the one input class that only ever surfaces if an admin happens to proactively call `list_reports`. In an active-but-not-checked-constantly community, a report can sit unseen indefinitely.

This also sharpens a residual risk `docs/SECURITY.md` already documents but doesn't mitigate: *"DM-originated content reports are visible only to super admins... the report is still recorded and retrievable, just only by super admins, until/unless a routing mechanism for DM-originated reports is added."* Today even a DM-originated report reaches super admins only if one manually thinks to check — there's no signal telling them one exists at all.

## Proposed approach

Smallest viable version, one PR, no new tool/schema/RBAC surface:

1. In the `reportContent` tool handler (`src/agent/tools.ts`), on a successful `createContentReport()` call, fire a super-admin DM alert reusing the exact same mechanism `audited()` already uses (`notifySuperAdmins`, `🔔`-prefixed `sendDirectMessage`, fire-and-forget with `.catch(logger.warn(...))`). Message includes: reporter name, conversation id, report id, and the reason (already capped at 500 chars, same content already stored — no new exposure).
[...]

PR #125, opened 2026-07-04
Give build worker room under pool contention (timeout 60→120), keep per-issue concurrency

swampratnz:
## Summary

Fixes the recurring "multiple proposals stuck at `needs-human`" problem. **Note the approach changed after review** — see below.

**Diagnosis (run 28703635677 + four failed attempts on #90):** the build worker's builds all draw on the **same shared Max pool**. When several are approved/retried at once they throttle each other's Claude turns, so each build blew past its **60-minute wall-clock** and was `cancelled` *mid-`npm test`*, producing no PR → `needs-human`. The build-retry loop then re-ran the failures, adding more parallel load — a contention spiral. #90 (a one-helper change) ran a full 59 min and was killed at the 60-min timeout.

**First attempt (reverted):** a single shared `concurrency.group` to force one-build-at-a-time. The automated review correctly caught that this is unsafe — a GitHub concurrency group keeps only **one** pending run, so a burst of approvals *cancels* all but the last-queued (`cancel-in-progress: false` only protects the running job, not the pending queue). A `cancelled` run isn't a `failure`, so the retry loop wouldn't rerun it and the verify/escalate step never runs — issues would sit at `status:approved` invisibly. Worse than the status quo.

**Shipped approach:**
- **Keep per-issue `concurrency`** (each issue its own group) so distinct issues never evict each other and same-issue re-triggers still queue — no silent drops.
- **Raise `timeout-minutes` 60 → 120** so a throttled/contended build finishes slowly rather than being killed mid-gate (and cancellations aren't retried). Still well under GitHub's 6 h job cap.
- Document why a shared serial group is the wrong tool, and note a **true FIFO lock** (a queue step the job polls) as the proper fix if bursts keep saturating the pool — deferred as its own change because a lock deployed wrong deadlocks every build.
- `CLAUDE.md` + `docs/PIPELINE.md` updated to match.

This is a **mitigation, not hard serialization**: it stops the 60-min-kill failure mode and removes all silent-drop risk, but a large enough simultaneous burst could still contend. Staggering approvals (or the future FIFO lock) covers that.

## Security / privacy impact

None. CI-orchestration only — no `src/`/`tests/`, no change to tool gating, RBAC, outbound filtering, secret redaction, or SQL scoping.

## How verified

- `python3 -c "import yaml"`: parses; `concurrency.group` is back to `pipeline-build-${{ github.event.issue.number }}` and `timeout-minutes` is 120.
- Root cause traced to a real run (build step `cancelled` at ~59 min with `npm test`/`node`/`claude` as terminated orphans — legitimate work killed by the wall-clock, not idle).
- No application code changed, so `typecheck`/`test`/`build` are unaffected.

Separately (not in this PR): tonight's stuck backlog is being drained by re-triggering the affected issues **one at a time** by hand, which sidesteps contention regardless of this change.

### DIFF
--- a/.github/workflows/pipeline-build.yml
+++ b/.github/workflows/pipeline-build.yml
@@ -28,14 +28,21 @@ on:
   issues:
     types: [labeled]
 
-# Per-issue concurrency: distinct approved issues build in PARALLEL (faster
-# backlog drain), while re-triggering the SAME issue queues instead of
-# double-running (kills duplicate PRs). GitHub can't express a hard "max N
-# concurrent", so throughput is governed by how many issues are released at
-# once; every run still draws on the shared Max pool, so release in small
-# batches to avoid exhausting the weekly token allowance.
+# SERIAL builds — one at a time across the whole repo (a single fixed group,
+# NOT keyed by issue). Why the change from per-issue-parallel: every build draws
+# on the SAME shared Max pool, and per-issue parallelism let N approved/retried
+# issues run at once, so each turn got rate-limited, every build blew past its
+# 60-minute wall-clock, and ALL of them escalated `needs-human` — then the
+# build-retry loop re-ran the failures, adding MORE parallel load: a contention
+# death spiral (observed 2026-07-04: five small proposals, incl. #90, all
+# cancelled at exactly 60 min with `npm test` still running). Serialising means
+# each build gets the full pool and finishes in time; the backlog drains one PR
+# at a time instead of failing all at once. `cancel-in-progress: false` queues
+# rather than cancelling, so a re-trigger of the same (or any) issue never
+# double-runs. Slower but reliable — the shared pool is the bottleneck, not CI
+# minutes.
 concurrency:
-  group: pipeline-build-${{ github.event.issue.number }}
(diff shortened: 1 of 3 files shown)


swampratnz:
Good catch — confirmed and fixed in c728227. You're right: a concurrency group holds only one pending run, so a shared `pipeline-build` group would cancel all but the last-queued build in a burst, and a `cancelled` run isn't a `failure` — so the retry loop wouldn't rerun it and the verify/escalate step never runs. Silent drops, worse than the `needs-human` pile-up.

Reverted to per-issue concurrency (no cross-eviction) and instead addressed the actual root cause — builds killed at the 60-min wall-clock while throttled on the shared Max pool — by raising `timeout-minutes` to 120 so a contended build finishes rather than being cancelled mid-gate. Documented why a shared serial group is the wrong tool, and noted a true FIFO lock (a queue step the job polls) as the proper fix if bursts keep saturating the pool — deferred deliberately, since a lock deployed wrong deadlocks every build. This is a mitigation, not hard serialization; staggering approvals covers the residual burst case meanwhile.

---

PR #272, opened 2026-07-08
Multi-choice polls + end_poll (Discord)

swampratnz:
Adds the poll capabilities that were missing when a poll needed changing to multi-choice.

## The Discord constraint (why there's no "edit poll")

Discord polls are **immutable** once posted — there is no edit endpoint, and discord.js exposes none. The question, options, duration, and single-vs-multi-choice setting are fixed at creation. The *only* poll mutation Discord supports is expiring a running poll early (`Poll#end`). So this PR adds the two things that are actually possible, and the tool descriptions are explicit that a live poll can't be converted.

## What's added

- **`create_poll` → `multiChoice` option** — post a poll where voters can pick more than one option. Defaults to `false` (single choice), so existing behaviour is unchanged; wired to Discord's `allow_multiselect`.
- **`end_poll` tool (admin)** — closes a running poll early: freezes the tally, stops further voting. It does **not** delete the poll or its votes. Guards against a message that isn't a poll, and is a no-op on an already-finalized poll (never double-ends).

## Answering the original questions

- *"Change the existing poll to multi-choice?"* — not possible as an edit. Use `end_poll` on the current poll (its results stay visible), then `create_poll … multiChoice: true` for the new one.
- *"Will updating wipe the choices?"* — ending a poll doesn't wipe anything (results stay). But because multi-choice requires a **new** poll, that new poll starts at **zero votes** — the old votes can't carry over. This is called out in the tool descriptions so the bot won't imply otherwise.

## Security / safety

Both reuse `create_poll`'s exact model: **admin tier**, **conversation-scope** validation (refuses channels the bot has never seen), **capability check** (Discord-only; WhatsApp cleanly reports unsupported), and **audit** logging. Not CONFIRM-gated — ending a poll is low-consequence and mirrors `create_poll`'s rate-capped-not-CONFIRM treatment. No new egress, data, or migration.

## Tests / validation

- `tsc`, `eslint`, `build`: clean.
- New tests: `create_poll` multiChoice threading (tools + adapter `allow_multiselect`); `end_poll` RBAC/scope/capability (**3 SECURITY**) + id threading; adapter `end_poll` ends / no-op-if-final / rejects-non-poll. `discordAdapter.test.ts` 51 pass; poll tests in `tools.test.ts` green. `security-floor.json` bumped (tools 76→79).
- Docs: SECURITY.md validated-targets list, CHANGELOG.


swampratnz:
Addressed the review (commit c1cebda):

- **`end_poll` now has a rate cap** — the main finding was correct: it had none. Added `POLL_END_RATE_LIMIT_PER_HOUR` (10/hr) + `reservePollEndSlot`, a *separate* per-conversation bucket from `create_poll`'s (so ending polls neither consumes nor is blocked by the create budget), enforced in `end_poll` with the same sliding-hour shape as `create_poll`. A hijacked/injected admin turn can no longer end live polls in a conversation unthrottled. Pinned by a new `SECURITY:` test; `security-floor.json` bumped 79→80.
- **CHANGELOG date nit** — folded the poll entry into the `2026-07-08` section (matching the commit date) and noted the rate cap, so the description and code now agree.

The `end_poll` (non-CONFIRM) rationale is now accurate: it *is* rate-capped, mirroring `create_poll`. On the reviewer's other non-blocking notes: CONFIRM-gating was considered and deliberately not applied (ending freezes future votes but deletes no data and hides no discussion, unlike `archive_thread`/`delete_message`) — consistent with the repo's consequence-based criteria.

PR #297, opened 2026-07-08
Fix generate_image: grok dropped GenerateImage, use /imagine

swampratnz:
`generate_image` has been failing on every call — found while running a proper end-to-end test on the server.

## Root cause

Grok removed the `GenerateImage` tool the bot invoked via `--tools GenerateImage`. A `--tools` allowlist naming a non-existent tool makes grok's agent build fail (`agent building failed: run_terminal_cmd …`). Not auth/quota/version (reproduced across stable 0.2.82/0.2.91 and alpha 0.2.93; bare grok works). Image generation is now grok's **`/imagine`** skill (built-in **`image_gen`** tool), which needs the full default toolset — so the old `--tools` allowlist lockdown is no longer usable.

## Fix + security model (all host-verified)

Invoke `/imagine`, and contain grok with a **kernel sandbox** rather than a tool filter (which proved unreliable — see below):

- **`grok --sandbox strict`** (landlock + seccomp on Linux): reads limited to the throwaway CWD + `~/.grok/` + system paths (**not** the bot's home/config), writes to CWD/`~/.grok/`/temp, child-process network blocked. This is the real containment.
- **No `--always-approve`**: headless grok cancels approval-gated tool calls (shell, file write) instead of running them.
- **`assertSandboxContains()` self-check**: once per process before the first generation, plant a sentinel outside the sandbox and confirm a sandboxed grok can't read it — **fail closed** (disable image gen) if the sandbox ever stops containing. Guards against a silent grok regression (the exact class of breakage that caused this bug); CI can't exercise a real grok, so the pure decision (`sandboxBreached`) is unit-tested.
- `--disable-web-search` removes the web tools; output handling (`parseSessionId`/`locateSessionImage`/`sniffImageType`) is unchanged.

### Why not a `--tools` allowlist or `--deny` list
Verified on the host that both fail: the image tool isn't `--tools`-selectable (allowlisting it breaks agent build), and `--deny bash`/`--disallowed-tools` did **not** block the shell (a `--deny` name that doesn't match grok's internal tool id fails **open** — a prompt still executed a shell command under `--always-approve`). And read tools are **auto-approved**, so without the sandbox an injected `/imagine` description could read `.env`/`~/.grok/auth.json`. Hence kernel sandbox + self-check.

## Verification (on the host, as the service user)

- Real image produced end-to-end under `--sandbox strict`.
- A prompt reading `/opt/community-agent/.env` → `Cancelled`, no secret leaked.
- A prompt ordering a shell write → `Cancelled`, no file written.
- `tsc`/`eslint`/`build` clean; `grokImage.test.ts` 7/7 (SECURITY test pins `--sandbox strict`, no `--always-approve`, no `--tools`; `sandboxBreached` unit-tested). `security-floor.json` unchanged.
- Docs: `docs/SECURITY.md §8` rewritten; CHANGELOG entry added.

No behaviour change unless `IMAGE_GEN_ENABLED=true` (it is on the live box — this restores the feature there once deployed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ loosely follows [Keep a Changelog](https://keepachangelog.com/). The agent's
 ## 2026-07-09
 
 ### Fixed
-- **Image generation works again** (`generate_image`). Grok removed the `GenerateImage` tool the bot was invoking, so every generation had been failing with an agent-build error; image generation is now grok's `/imagine` skill. Reworked the integration to call it and re-established the security lockdown without the old `--tools` allowlist (which is no longer possible): the subprocess runs with **no `--always-approve`** — so headless grok cancels any shell/file/subagent/MCP tool call instead of running it — plus explicit `--deny` rules for those tools and `--disable-web-search`. Verified end-to-end on the host (a real image is produced; a prompt ordering a shell command is cancelled and writes nothing). No behaviour change unless `IMAGE_GEN_ENABLED=true`. See docs/SECURITY.md §8.
+- **Image generation works again** (`generate_image`). Grok removed the `GenerateImage` tool the bot was invoking, so every generation had been failing with an agent-build error; image generation is now grok's `/imagine` skill. Reworked the integration to call it and re-established the security lockdown without the old `--tools` allowlist (no longer possible), using a **kernel-enforced `--sandbox strict`** (the agent can't read the bot's secrets or reach the network of a child process) plus **no `--always-approve`** and `--disable-web-search`. Verified end-to-end on the host: a real image is produced, while a prompt trying to read `/opt/community-agent/.env` or run a shell command is cancelled with nothing leaked or written. No behaviour change unless `IMAGE_GEN_ENABLED=true`. See docs/SECURITY.md §8.
 
 ## 2026-07-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ loosely follows [Keep a Changelog](https://keepachangelog.com/). The agent's
 `whats_new` tool reads this file, so keep entries user-legible and add a new
 `##` dated section (or version) as part of each release.
 
+## 2026-07-09
+
+### Fixed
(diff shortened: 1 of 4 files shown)


swampratnz:
Both review concerns were correct — I verified them on the host and reworked the fix around a **kernel sandbox** instead of a tool deny-list (commit 4201fbf):

1. **Deny-list was unreliable (concern #1).** With `--always-approve`, a prompt still executed a shell command and wrote a file *despite* `--deny bash`/`--disallowed-tools bash,run_terminal_cmd` — the name doesn't match grok's internal tool id, so it fails **open**. Confirmed.
2. **Read tools are auto-approved (concern #2).** Without `--always-approve`, a prompt read `/etc/hostname` and returned it — so an injected `/imagine` description could have exfiltrated the bot's secrets (`.env`, `~/.grok/auth.json`). Confirmed.

**New control — `--sandbox strict`** (kernel-enforced, landlock + seccomp on Linux), which doesn't depend on knowing tool names:
- reads limited to the throwaway CWD + `~/.grok/` + system paths (**not** the bot's home/config), writes to CWD/`~/.grok/`/temp, child-process network blocked.
- **Verified on the host:** `/imagine` under `--sandbox strict` still generates an image (grok reads its own auth, calls the image API, writes the image); a prompt reading `/opt/community-agent/.env` is **`Cancelled`, nothing leaked**; a shell-write prompt is still cancelled (no `--always-approve`).

Dropped the unreliable `--deny` flags. Updated the `buildGrokArgs` SECURITY test to pin `--sandbox strict` (+ no `--always-approve`, no `--tools`), `docs/SECURITY.md §8`, and CHANGELOG. `security-floor.json` unchanged (still a rewrite, not add/remove). `tsc`/`eslint`/`build`/tests green.

swampratnz:
Addressed both points (commit cfc90fa): (1) corrected the PR description — the shipped control is `--sandbox strict` (kernel-enforced) + no `--always-approve`, not a `--deny` list (the `--deny`/`--disallowed-tools` approach was dropped precisely because it failed open in host testing). (2) Added a per-process runtime self-check (`assertSandboxContains`) that plants a sentinel outside the sandbox and confirms a sandboxed grok can't read it before the first generation, failing **closed** if the sandbox ever stops containing — so a silent grok regression can't quietly reopen arbitrary-file-read. The pure decision (`sandboxBreached`) is unit-tested since CI has no grok binary.

PR #418, opened 2026-07-12
ci: self-healing changelog autofill (draft-PR companion to the coverage tripwire)

swampratnz:
Closes the loop on the changelog-coverage tripwire. Today `changelog-coverage.yml` only opens a **tracking issue** listing merged user-facing PRs missing a `CHANGELOG.md` entry — a human still has to write the prose. This adds **`changelog-autofill.yml`**, which runs the *same* coverage check and, when there's a gap, has Claude draft house-style entries and open **one standing PR** you just review-and-merge.

### How it works
- Runs daily 30 min after the coverage check (+ `workflow_dispatch`), **inert** until `CLAUDE_CODE_OAUTH_TOKEN` is set.
- Recomputes gaps via `scripts/check-changelog-coverage.mjs`; if none, exits.
- **One standing PR** on fixed branch `chore/changelog-autofill`. If an autofill PR is already open it **skips** — never force-pushes over your review, never opens a second. When you merge/close it, the next run drafts a fresh one for whatever gap remains.
- The agent reads each flagged PR, **skips genuinely-internal ones**, writes an entry under the correct dated section, runs `prettier --check` + the `whats_new` parser test (`tests/changelog.test.ts`), pushes, and opens the PR.

### Guardrails (mirrors `pipeline-pr-conflict.yml`'s posture)
- Agent may touch **only `CHANGELOG.md`** — prompt-stated, then a deterministic **scope tripwire** reads the opened PR's diff and labels `needs-human` + comments if anything else changed.
- **Never merges** — a human does; the agent's `gh` has no merge/close verb.
- PR opened via the action's **App token** so the PR's required CI actually runs (a `GITHUB_TOKEN`-opened PR triggers no workflows).
- `GITHUB_TOKEN` kept off the agent step; `persist-credentials: false`; `git push origin HEAD` is the only push form.

`changelog-coverage.yml`'s tracking-issue body now points readers at the `chore/changelog-autofill` PR.

### Verify
- Both workflow YAMLs parse.
- No behaviour change until merged + `CLAUDE_CODE_OAUTH_TOKEN` present; safe to merge inert.

### DIFF
--- a/.github/workflows/changelog-autofill.yml
+++ b/.github/workflows/changelog-autofill.yml
@@ -241,21 +241,26 @@ jobs:
                `npx prettier --check CHANGELOG.md` (must pass).
             5. `node --import tsx --test tests/changelog.test.ts` — the parser the
                `whats_new` tool relies on. It must pass (0 failures).
-            6. Confirm ONLY CHANGELOG.md changed: `git status --porcelain` must
-               list nothing but ` M CHANGELOG.md`. If anything else is dirty, undo
-               it — you may edit only CHANGELOG.md.
-            7. Commit: `git add CHANGELOG.md` then a `git commit` whose message is
+            6. Confirm CHANGELOG.md is the ONLY tracked change:
+               `git status --porcelain -- CHANGELOG.md` must show exactly
+               ` M CHANGELOG.md`. (The list files `gaps.md`/`prs.json` in the repo
+               root are gitignored CI scratch — ignore them; do NOT try to delete
+               them and do NOT run `git add -A` or `git commit -a`. You may stage
+               only CHANGELOG.md.)
+            7. Commit: `git add CHANGELOG.md` (that exact path — nothing else)
+               then a `git commit` whose message is
                `docs: backfill CHANGELOG.md for recently-merged PRs`.
             8. Push with EXACTLY: `git push origin HEAD` (no other push form,
                never force).
-            9. Open the PR with `gh pr create --base main --head
-               ${{ env.AUTOFILL_BRANCH }} --label ${{ env.AUTOFILL_LABEL }}
-               --label documentation`, titled
-               `docs: backfill CHANGELOG.md for recently-merged PRs`. In the body:
-               list each PR you documented (with its `#NNN`), name any flagged PR
-               you deliberately SKIPPED as internal and why (one line each), and
-               state that this was auto-drafted and needs human review before
-               merge. Do NOT merge or close the PR — a human merges.
+            9. Open the PR by running EXACTLY this command, with `--base main
+               --head chore/changelog-autofill` first and in that order (the tool
+               allowlist pins that literal prefix, so any other order or head is
+               rejected):
+               `gh pr create --base main --head chore/changelog-autofill --label changelog-autofill --label documentation --title "docs: backfill CHANGELOG.md for recently-merged PRs" --body "<body>"`.
+               In the body: list each PR you documented (with its `#NNN`), name any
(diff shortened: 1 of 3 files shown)


swampratnz:
Addressed the automated review (commit bd4aae7):

- **Cleanup nit** — dropped the dead `head_before` output (the `id: mark` step is gone).
- **Trust-surface** — hardened rather than just accepted:
  - Removed `gh pr list` from the agent's allowlist — it works from `gaps.md` + `gh pr view/diff <N>`, so it can no longer enumerate arbitrary PRs.
  - The scope gate now **fails closed**: an out-of-scope diff **closes the PR and deletes the branch** instead of only labelling `needs-human`, so an injected non-`CHANGELOG.md` change can never sit in a PR a reviewer might rubber-stamp.
  - Documented the risk-acceptance in the header, enumerating the bounds: no `gh pr list`, no `GITHUB_TOKEN` on the agent, fail-closed scope gate, App token can't merge / push `main` / touch `workflows`, and a human merges every PR — so there's no path from an injected historical PR to production without a human merge.

The residual (point 1) is a one-time risk-acceptance for feeding contributor-authored PR content to a write-capable agent. Flagging for @swampratnz sign-off.

swampratnz:
Addressed the second review (commit 664b750):

- **High — `gh pr comment` side-effect vector:** **removed `gh pr comment` from the agent allowlist entirely.** No prompt step used it, and its target number can't be matcher-pinned, so it was exactly the uncontained vector you describe (a forged `pipeline-pr-*` marker on an unrelated PR would be invisible to my diff-only scope gate). All commenting is now done only by the deterministic `GH_TOKEN` steps. This also brings the agent in line with `pipeline-build.yml`, which likewise withholds `gh pr comment` from its build agent. The agent's `gh` surface is now **`create`/`diff`/`view` only**.
- **Minor — `gh pr create` wildcard:** kept (its `--title`/`--body` are dynamic, so it can't be exact-matched — same as the build worker), but documented the containment explicitly in both the header and the tool comment: the agent can only push `AUTOFILL_BRANCH` (single pinned push form + `checkout`/`branch`/`switch` withheld) and the App token is installation-scoped, so an injected `--head`/`--repo` can at worst open a **visible, human-gated, unmergeable** PR from a pre-existing branch — never a comment side-effect, never cross-repo, never a merge.

Thanks — the comment-vector catch was a real gap in the containment story.

swampratnz:
Addressed the third review (commit 3baa87f):

- **Real bug — impossible clean-tree check on `gaps.md`/`prs.json`:** fixed properly.
  - **gitignored** `/gaps.md` and `/prs.json` (mirroring the existing `/var/` scratch precedent) so they never register as dirty and a stray `git add -A` can't stage them.
  - Reworded step 6 to `git status --porcelain -- CHANGELOG.md`, and told the agent the scratch files are gitignored — don't delete them, don't use `git add -A`/`git commit -a`.
  - **Pinned `Bash(git add CHANGELOG.md)`** (was `git add:*`), so `git add -A` is now literally impossible — the scope-widening path you flagged is closed at the tool layer, not just by instruction.
- **Optional hardening — taken:** pinned `Bash(gh pr create --base main --head chore/changelog-autofill:*)` (was `gh pr create:*`), matching step 9's literal prefix, so an injected `--head`/`--repo` is denied. Step 9 now specifies the exact command and flag order.

The agent's write surface is now: edit CHANGELOG.md, `git add CHANGELOG.md` (exact), commit, `git push origin HEAD` (exact), and `gh pr create` with a pinned `--base main --head chore/changelog-autofill` prefix — nothing else can stage a file, push a branch, comment, or open a PR from another head.

PR #421, opened 2026-07-12
feat: super-admin dev-team tools (dispatch/status/result + completion DM)

swampratnz:
## What

Super-admin-only tools that let Dave drive a remote **dev-team** build service over the tailnet — dispatch a job (assess/deliver), check status, fetch the result — and **proactively DM the requester when a ~20-min run finishes**.

### Tools (all super-admin, TEXT-only so they work on Discord and WhatsApp)
- **`dev_team_dispatch`** `{mode, repo, title?, description?, budget_usd?}` — POSTs a job to the service, inserts a durable completion-DM watch, and replies with the job id + "~20 min; I'll DM you when it's done." `mode:"deliver"` (which makes changes / opens a PR) is **CONFIRM-gated**; `assess` (read-only) runs without confirmation.
- **`dev_team_status`** `{id?}` — one job's status, or a recent-jobs listing when no id.
- **`dev_team_result`** `{id}` — the assessment verdict (classification + executive summary + top of report) or the delivery outcome, capped for chat with a pointer to the dashboard for the full artifact.

## Security posture

**Two-layer super-admin gate.** The tools are added to `SUPER_ADMIN_TOOLS` (so a lower tier never even sees them in its tool list), and each handler re-asserts `assertAtLeast(caller.role, 'super_admin', …)` before any side effect — defence in depth identical to `suggest_issue` / `redeploy_bot`.

**deliver-CONFIRM.** `deliver` registers a pending action via the `pendingActions` flow and returns a CONFIRM prompt; the dispatch only runs on the super admin's own out-of-band CONFIRM reply, which never passes through the model — so an injected turn can request a delivery but can never complete one.

**Tailnet http rationale.** `DEV_TEAM_ENDPOINT_URL` deliberately allows `http://`. The service is a tailnet-internal endpoint reached only over WireGuard-encrypted, device-authenticated Tailscale, which already provides transport confidentiality/authentication. Unlike the public-internet `DOCS_INGEST`/`STATUS_CHECK` URLs there is intentionally **no** `.startsWith('https://')` guard (commented in `config.ts`). A `.refine` still fails fast at startup if the feature is enabled without both the URL and the token.

**Token handling.** `DEV_TEAM_AUTH_TOKEN` is sent only as a `Authorization: Bearer` header; the client never logs it and never puts it in a thrown error (which carries status + a 200-char-capped body only). It is also added to `runtimeSecrets()` so the outbound DLP layer redacts it belt-and-braces, and any echoed service text surfaced to chat is scrubbed + capped.

**Durable completion DM.** A `dev_team_watches` table records who dispatched each job; a poller (default 1-min cadence, gated on `DEV_TEAM_ENABLED`, unref'd) reads unnotified watches, checks each job's status over the tailnet, and on a terminal state DMs the requester on their own platform then stamps `notified_at` — restart-safe and **send-once** (primary key + notified guard, and a failed DM leaves the row for the next tick to retry).

## Client

Typed `dispatchJob` / `jobStatus` / `jobResult` / `listJobs` coding to the frozen HTTP contract, mirroring the injectable-fetch pattern in `anthropicStatus.ts`: bearer header, `AbortSignal.timeout(15s)`, throw-on-non-2xx with a capped body, `fetchImpl` injectable so tests never hit the network.

## Tests

RBAC gating (members/guests/admins never get the tools; super-admins do), handler super-admin re-checks, deliver-CONFIRM registers a pending action without dispatching, disabled-config friendly message, config parse + refine, client bearer/cap/token-secrecy, `runtimeSecrets()` redaction, and watch repository + poller (DMs on terminal state, marks notified, no double-send, skips when disabled). `security-floor.json` bumped via `test:security:fix`.

## Gates

`typecheck` / `lint` / `format:check` / `build` green. `test` and `test:security` green on Linux CI (all SECURITY tests pass; DB-backed cases skip cleanly without a local Postgres).


swampratnz:
Addressed the review (commit 2c1c420):

- **Prompt-injection quarantine — fixed.** Every service-originated free-text field (`error`, progress `role/stage/message`, `classification`, `executive_summary`, `report_markdown`) is now wrapped in the same `untrusted()` quarantine used for web-search and recalled-message content before it reaches the model — labelled as data, newlines/angle-brackets stripped, so injected text can never start a fresh model-visible line. Short identifier fields (`id`/`mode`/`repo`/`state`/timestamps) get a bracket/newline neutralizer (`devTeamField`) since the contract's slugs/enums weren't client-enforced. 3 new `SECURITY:` tests assert the quarantine end-to-end (an injected `IGNORE PREVIOUS INSTRUCTIONS\n… <system>` payload in every field); `devTeamTools.test.ts` floor 2→5.
- **Minor note (no CONFIRM/rate-cap on assess) — deliberate, now documented here:** `assess` is read-only on the dev-team side and super-admin-only on this side; cost is bounded per-job via `budget_usd`, and the dispatch service itself enforces single-flight + a queue cap of 16, so unbounded parallel spend isn't reachable. `deliver` (the side-effecting mode) stays CONFIRM-gated. If a daily cap on assess dispatches proves wanted in practice, it can be added the same way as `GITHUB_ISSUE_DAILY_LIMIT`.

swampratnz:
Addressed the second review round (commit 1a6747e):

1. **Privacy — `dev_team_watches` wired into purge.** `purgeSingleIdentity` now deletes the purged identity's watch rows (`requester_platform`/`requester_user_id` keyed, same as every sibling table), counted in the purge total. SECURITY test mirrors the `knowledge_gaps` purge test (target's rows gone, another requester's untouched); `repository.test.ts` floor 53→54.
2. **Rate cap on dispatch.** New `DEV_TEAM_DAILY_LIMIT` (default **10**/day per super admin, `0`=unlimited) enforced by `reserveDevTeamDispatchDaily` — the same calendar-day shape as `GITHUB_ISSUE_DAILY_LIMIT`/`IMAGE_GEN_DAILY_LIMIT`. Reserved **inside the dispatch closure**: after `deliver`'s CONFIRM (a denied confirmation consumes nothing) and before the POST (a failed POST is deliberately not refunded, so induced-failure retries can't bypass it). Documented in `.env.example`. SECURITY test on the reserve function; `devTeamTools.test.ts` floor 5→6.
3. **Completion-DM neutralization.** `devTeamField` moved to `src/devTeam/client.ts` (single shared neutralizer — no drift between the chat formatters and the DM), and `formatDevTeamCompletionDm` now runs `watch.mode`/`jobId`/`repo` and the capped service error through it. SECURITY test proves an injected `\n<...>` payload can't add lines or brackets to the unprompted DM; `devTeamWatch.test.ts` floor 4→5.

All local gates green (typecheck / lint / format / build / security-manifest static check; new tests 15 pass 0 fail).

swampratnz:
Addressed review round 3 (commit HEAD):

- **Orphaned-job failure mode — fixed as partial success.** The watch insert after a successful POST is now best-effort: on failure the reply carries the **job id**, an explicit **"NO completion DM will come"** caveat, and a pointer at `dev_team_status`; the error is logged with the job id. It can never read as "Failed to dispatch" (which invited a cost-doubling retry). New SECURITY test in its own file (`devTeamDispatchPartial.test.ts` — it must mock `repository.js`, and module mocks are process-wide) proves the POST-succeeded/insert-failed case surfaces the id and never reports failure. The daily-cap reservation correctly still counts this dispatch (the job IS running).
- **Non-blocking nit — taken anyway:** the deliver-CONFIRM prompt and dispatch success text now run `repo`/`mode`/`job.id` through `devTeamField` before splicing.

swampratnz:
Addressed round 4: unrolled the looped admin-rejection tests into **3 static `test('SECURITY: …')` declarations** (one shared helper, no behaviour change) and set `tools.test.ts`'s manifest entry to the **true static count 126**. The gate's static scan now counts exactly what runs — dropping any one of the three would fail `security-invariants`.

PR #518, opened 2026-07-14
test: fix agentCoreFailureFallbacks* exit-7 crash (onnxruntime teardown)

swampratnz:
## Summary

Fixes the intermittent `security-invariants` CI crash (`tsx --test` exit code 7,
no failing assertion) that has sporadically blocked otherwise-green PRs.

**Root cause (corrected after on-CI diagnosis).** The four
`agentCoreFailureFallbacks*` test files drove the real `runAgentTurn`, whose
memory-recall step (`searchMemory` → `embed`) loads the `onnxruntime-node`
embedding model. In the no-Postgres `security-invariants` job, under full-suite
CPU load, that **native inference session's teardown could race the child test
process's exit and abort it via Node's fatal-error path**: exit 7, no failing
assertion, log output truncated mid-flush. Crucially it is **uninterceptable by
any JS `unhandledRejection`/`uncaughtException` handler** — I confirmed on CI
that even a `process.prependListener`'d handler (ahead of node:test's own) never
fired. That's why it read as a mysterious infra failure, was near-deterministic
on the contended CI runner, and did not reproduce across ~19 local full-gate
runs (the sandbox's different load never hit the teardown race).

An earlier revision of this PR assumed the cause was a fire-and-forget DB
`ECONNREFUSED` rejection and stubbed the pool. That removed the connection noise
but **not** the crash — the on-CI diagnostic rounds (since squashed out) proved
the fault survived with zero `ECONNREFUSED`, pointing to the native path above.

## The fix

- **Mock `searchMemory` in all four files** so the embedding model never loads —
  removing the native component entirely. These tests never assert on recalled
  memory (`query()` is mocked, so prompt content is moot), so this is behaviour-
  preserving for what they actually verify.
- **Also stub `src/storage/db.js`'s `pool`** with a resolve-empty double
  (installed before the `repository.js` import so `repository.ts` binds it at
  module-evaluation time), so the remaining light DB reads
  (`getCodeAnswersPolicy`, `getClaudeSession`, …) open no socket and emit no
  `ECONNREFUSED` noise. Resolve-empty (never throwing) so it can surface no
  rejection of its own.

## Security / privacy impact

None. Test-only change. No `SECURITY:`-prefixed tests were added or removed, so
`tests/security-floor.json` is unchanged (manifest total stays 647 across 87
files). No RBAC / tool-gating / outbound-filtering / secret-redaction /
SQL-scoping / confirm-flow code is touched. The `SECURITY:`-tagged fail-open
assertions in these files exercise the same substitution paths and are
unaffected by the mocks.

## How verified

- `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run build`
  — all green.
- Against a dead DB port, the four files now load **no** embedding model
  (0 "Loading embedding model") and open **no** socket (0 `ECONNREFUSED`), and
  pass.
- `npm run test:security` runs clean across repeated runs (exit 0, 647/647)
  where the same gate previously crashed exit-7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,27 @@ is a NZ community, and the CI that opens most PRs runs in UTC (a day behind NZ
 for anything after ~noon NZST/NZDT). Get today's date with
 `TZ='Pacific/Auckland' date +%F` rather than a bare `date`.
 
+## 2026-07-15
+
+### Fixed
+- **Intermittent `security-invariants` CI crash (exit 7)**: the four
+  `agentCoreFailureFallbacks*` test files drove the real `runAgentTurn`, whose
+  memory-recall step (`searchMemory` → `embed`) loads the `onnxruntime-node`
+  embedding model. In the no-Postgres `security-invariants` job, under
+  full-suite CPU load, that native inference session's teardown could race the
+  child test process's exit and abort it via Node's fatal-error path —
+  `tsx --test` exit 7, no failing assertion, output truncated mid-flush, and
+  (confirmed) uninterceptable by any JS `unhandledRejection`/`uncaughtException`
+  handler, even a `prependListener`'d one. The four files now mock
+  `searchMemory` so the embedding model never loads (they never assert on
+  recalled memory — `query()` is mocked, so prompt content is moot), removing
+  the fault entirely. They additionally stub `src/storage/db.js`'s `pool` with
+  a resolve-empty double (installed before the `repository.js` import so the
+  binding is fixed at module-evaluation time) so the remaining light DB reads
+  open no socket and emit no `ECONNREFUSED` noise. Test-only change; no
+  `SECURITY:` tests added or removed, so `tests/security-floor.json` is
+  unchanged.
+
 ## 2026-07-14
 
 ### Added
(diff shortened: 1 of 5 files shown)


swampratnz:
Closing without merging — the exit-7 flake this PR set out to fix is not fixable from here, and the maintainer has opted to accept it as pre-existing infra noise (it already clears via `ci-retry`/autofix reruns; all recently-merged PRs passed this gate eventually).

Diagnosis for the record (in case it recurs), from four on-CI diagnostic rounds (since reverted):

- The crash is an **abrupt child-process exit 7 with no JS-observable error**. Confirmed on CI that neither `process.on('uncaughtException')`/`unhandledRejection`, a `prependListener`'d variant, **nor `uncaughtExceptionMonitor`** (which fires even under node:test's `setUncaughtExceptionCaptureCallback`) ever fired. So it is **not** a JS exception or promise rejection.
- Output is always truncated mid-`pino` flush, immediately after these files' `getLanguagePreference`-rejects test, then exit 7. It is flaky/near-deterministic on the contended CI runner and does **not** reproduce in a sandbox across ~25 full-gate runs (isolation, high `--test-concurrency`, CPU-loaded).
- **Ruled out:** the DB `ECONNREFUSED` race (a `db.js` pool stub removed all sockets — crash persisted); `onnxruntime-node` teardown (mocking `searchMemory` stopped the model from ever loading — crash persisted); any JS-level error handler.
- Remaining suspicion: a teardown-layer fault in `tsx` + `node:test` + `--experimental-test-module-mocks` (module-mock restoration or async log-stream flush), which test-code changes can't intercept.

The `searchMemory`-mock + `db.js`-pool-stub changes on this branch are legitimate test-isolation improvements (no real socket/onnxruntime I/O, faster) but do **not** fix the flake, so this PR is closed rather than merged under a misleading title.

---

Issue #675, opened 2026-07-23
Deflake the reply-budget count assertions in the shortcut-router tests (shared-table interference)


swampratnz:
## Problem (measured)

Two sibling router tests fail intermittently in CI with the identical signature — a reply-budget count one higher than expected:

- `tests/repeatMaxTurnsShortcutRouter.test.ts:287` — `3 !== 2`, "both the real failure reply and the repeat-shortcut reply must be recorded and counted toward the daily budget" (PR #673 CI, 2026-07-23 00:16 UTC).
- `tests/repeatQuestionShortcutRouter.test.ts:381` — `3 !== 2`, same wording (PR #672 CI, 2026-07-23 09:30 UTC, on a **changelog-only** diff, after already flaking on the prior head).

This is the same shared-table interference class the repo has hit repeatedly (`retryOnSharedTableInterference` in `tests/repository.test.ts` exists for it, and #661 instrumented the knowledge-gap variant): test files run as parallel processes against ONE Postgres, and any assertion that counts rows over a shared table (`interactions`) with a predicate looser than the test's own run-unique identifiers can absorb another file's concurrent insert. Observed cost: it has burned the ci-retry rerun AND a manual re-kick on multiple PRs this week — one of the two biggest remaining sources of merge latency now that the auto-merge lane works.

## Proposed fix (test-only)

For each of the two assertions (and any sibling in those files using the same shape):

1. **Prefer scoping over retrying**: if the underlying count query used by the assertion can be filtered by the test's own run-unique `user_id`/`conversation_id`, do that — deterministic, no absorb logic needed. (The production code path counts per-user daily replies; if the test asserts via a raw table count instead of the per-user helper, switch it to the scoped helper.)
2. Where the assertion deliberately exercises an unscoped/aggregate read, use the established `retryOnSharedTableInterference`-style bounded retry with a `console.warn` per absorbed attempt — matching `tests/repository.test.ts`'s existing convention — or a watermark (`id > max(id)` at test start) so only the test's own rows count.
3. A genuine regression (shortcut reply not recorded) must still fail after bounded retries.

## Approved scope (adversarial review, 2026-07-23)

**Bounded to the shared-table row-count/read interference class only.** Explicitly OUT of scope for this PR (each is a separate issue if pursued):
- The `#422` / `repository.test.ts:2393` embedding-similarity flake (`${RUN}` marker leaking into embedded `content`) — a *different root cause*, its own issue.
[...]

Issue #432, opened 2026-07-12
Low-rated-knowledge caveat only ever reaches members via the shortcut path — the normal knowledge_search path (most real questions) shows no warning


swampratnz:
## Problem statement

**Who it helps**: members asking anything that isn't a near-exact FAQ match — i.e. most real questions — who would otherwise be served a knowledge entry the community has already flagged as unhelpful, with zero warning. Admins get the correct signal today (post-#411); members served through the dominant answer path do not.

**Evidence, verified directly in `origin/main`:**

- `formatKnowledgeCitationNote` (`src/agent/tools.ts:265-289`) takes a `lowRatedCaveat` boolean (issue #337) that, when true, appends `KNOWLEDGE_LOW_RATED_CAVEAT_TEXT` to a hit's note. Its own doc comment (`tools.ts:251-256`) states this is "deliberately opt-in per call site: only the member `sendKnowledgeShortcut` path in router.ts ever computes and passes `true`."
- `src/router.ts:722-740` (`sendKnowledgeShortcut`) is the *only* call site that computes it — via `isKnowledgeLowRated(hit.id, minUnhelpful)` (`repository.ts:4144`) — gated behind `KNOWLEDGE_LOW_RATED_CAVEAT_MIN_UNHELPFUL > 0`.
- `formatKnowledgeSearchResults` (`tools.ts:336-395`), which renders the tool-result text the model reads for every ordinary `knowledge_search` call, hardcodes the argument: `formatKnowledgeCitationNote(h, staleDays, false, maxAgeDays)` (`tools.ts:390`) — always `false`, regardless of config or how low-rated the entry actually is.
- The shortcut only fires above `KNOWLEDGE_SHORTCUT_THRESHOLD` (default 0.9 cosine, near-exact match) — a narrow slice of traffic (issue #163's own framing). Everything else — the common case, where a member's phrasing doesn't hit that ceiling — goes through the model-mediated `knowledge_search` tool call, which this proposal's evidence shows carries no low-rated signal at all.
- `systemPrompt.ts` gives the model no separate guideline to relay an unhelpful-rating warning either (grep confirms only `rate_answer`-trigger guidance exists), so there's no fallback path by which the model could surface this on its own.

[...]

PR #683, opened 2026-07-25
knowledge_search: degrade gracefully when a secondary lookup fails

swampratnz:
## Summary

Two fixes found during a review of the repo for security issues, formatting, and bugs.

**1. `knowledge_search` discarded a good answer when an advisory lookup failed.** The conflict-badge check and the below-floor lexical fallback were the only two unguarded database awaits in the handler. Their immediate neighbours all already degrade — `recordKnowledgeRetrieval` and `recordKnowledgeGap` are fire-and-forget with an inline catch, and `areKnowledgeEntriesLowRated` falls back to an empty set — and `router.ts`'s `sendKnowledgeShortcut` states the intended rule outright for the sibling path: *"a lookup failure falls back to `false` rather than blocking the reply."*

By the time either call runs, `hits` has already been fetched successfully. A transient failure (pool exhaustion, statement timeout — the trigram scan in the lexical fallback is the query most likely to hit one) therefore replaced a complete answer with an `isError` tool result carrying a raw driver message. Both now catch and drop only the supplementary signal.

**2. Three tests reached a live database despite comments asserting they did not**, so they failed for anyone without a local Postgres instead of skipping cleanly, contrary to the contract in CLAUDE.md:

- `docsIngest.test.ts` claimed the first three fetch-failure-batching tests ran before any DB call. That holds only for the all-fetches-fail case, where `seen` stays empty and the run returns before the prune's `listGlobalKnowledgeTitlesByProvenance`. The other two have a successful fetch and so reach both that query and `syncGlobalKnowledgeByProvenance`; they now carry `{ skip }` like every other DB-touching test in the file. The diff there is mostly re-indentation from Prettier rewrapping the two `test(...)` calls, matching how the file already formats its skipped tests.
- `agentCoreKnowledgeEntryId.test.ts` mocked only `searchKnowledge`, but its `'below-floor query'` fixture is precisely what reaches the lexical fallback. `searchKnowledgeLexical` is now stubbed too, so no fixture depends on a reachable DB.

## Security / privacy impact

No change to auth, tool gating, tier derivation, data-access scope, or stored data. No tool moved tier and no new tool was added.

One small improvement to outbound exposure: the unguarded awaits let a raw `pg` error message become the tool result, which the MCP layer returns verbatim into model context — carrying internal detail such as host:port and relation names that the outbound filter does not treat as a secret (it redacts exact known values and `postgres://` DSNs, but not those). Both paths now log the error server-side and return normal results instead.

`npm run test:security` is unchanged at 876 across 110 files — no `SECURITY:` test was added or removed, so `tests/security-floor.json` needs no update.

## How verified

Full gate green locally, no local Postgres (`DATABASE_URL` unset):

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm test` — **2419 tests, 0 fail** (1821 pass, 598 skip). Before this change: 3 consistent failures, reproduced across 4 runs.
- `npm run test:security` — passes, 876 tests, manifest matches
- `npm run build` — clean

The DB-backed paths themselves are unverified locally and will get their real coverage from CI's `pgvector/pgvector:pg16` service container, including the two `docsIngest` tests this PR marks `{ skip }` — they still run there.

Separately noted during the review but **not** changed here, as out of scope for this diff: `cloudAdapter.ts` compares the webhook verify token with `===` (non-constant-time) while the HMAC path on the same file correctly uses `timingSafeEqual`. Low severity — a one-off setup handshake where network jitter swamps the signal — but happy to follow up if wanted.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,16 @@ for anything after ~noon NZST/NZDT). Get today's date with
   nothing to confirm. Off by default; byte-identical behaviour when the flag
   is unset.
 
+### Fixed
+- **`knowledge_search` no longer throws away a good answer when a secondary
+  lookup fails.** The advisory conflict-badge check and the below-floor
+  lexical fallback were the only two unguarded database calls in the handler,
+  so a transient failure in either replaced results the bot had already
+  fetched successfully with a raw error. Both now degrade the same way their
+  neighbours (the retrieval counter, the low-rated caveat, the gap recorder)
+  already did: the supplementary signal is dropped and the search results are
+  still returned.
+
 ## 2026-07-23
 
 ### Added
--- a/src/agent/tools.ts
+++ b/src/agent/tools.ts
@@ -2709,7 +2709,16 @@ export function buildToolServer(
       // scoped, LIMIT-1 self-join — never the full-table audit
       // `listKnowledgeConflictCandidates` runs. Skipped entirely below 2
       // ids, matching hasConflictAmongIds' own zero-query short-circuit.
-      const hasConflict = relevantIds.length >= 2 ? await hasConflictAmongIds(relevantIds) : false;
+      // Fail-safe like the low-rated caveat below: the conflict note is a
+      // purely advisory badge, so a lookup failure must degrade to "no note"
+      // rather than discarding the hits we already fetched successfully.
+      const hasConflict =
+        relevantIds.length >= 2
+          ? await hasConflictAmongIds(relevantIds).catch((err) => {
+              logger.warn({ err }, 'Knowledge conflict check failed; omitting the conflict note');
+              return false;
+            })
+          : false;
       // Member-facing low-rated-answer caveat (issue #432) — the display-side
       // counterpart to #337's shortcut-only caveat: this is the dominant
       // answer path (below the shortcut's 0.9-cosine ceiling), so gating and
@@ -2737,9 +2746,16 @@ export function buildToolServer(
       // output is byte-identical to before issue #362 for the common case.
       let lexicalHits: Awaited<ReturnType<typeof searchKnowledgeLexical>> = [];
       if (hits.length > 0 && relevantIds.length === 0) {
+        // Fail-safe, same reasoning as the conflict/low-rated lookups above:
+        // this is a supplementary second attempt on the below-floor branch,
+        // so a failure here must degrade to "no lexical hits" and still show
+        // the semantic results, never replace them with a raw DB error.
         lexicalHits = await searchKnowledgeLexical(args.query, {
           platform: caller.platform,
           conversationId: caller.conversationId,
+        }).catch((err) => {
+          logger.warn({ err }, 'Knowledge lexical fallback failed; returning semantic results only');
+          return [];
         });
       }
       if (lexicalHits.length > 0) {
--- a/tests/agentCoreKnowledgeEntryId.test.ts
+++ b/tests/agentCoreKnowledgeEntryId.test.ts
@@ -7,14 +7,17 @@ import type { KnowledgeSearchHit } from '../src/storage/repository.js';
 // config.ts validates env at import time — provide a dummy environment
 // before importing anything that (transitively) loads it, matching the
 // convention in tests/agentCoreLanguagePreference.test.ts. DATABASE_URL is
-// unreachable; the only repository export this file overrides is
-// searchKnowledge, so it can simulate a knowledge_search tool call's hits
-// deterministically without a real DB or embedding model — every other
-// repository call the exercised knowledge_search handler makes
-// (recordKnowledgeRetrieval, hasConflictAmongIds) is fire-and-forget/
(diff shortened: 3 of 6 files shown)


swampratnz:
Addressed in `0348ab9`.

## Review point: missing `SECURITY:` coverage

Agreed, and thanks for pointing at the precedent — `tests/knowledgeSearchFailSafe.test.ts` adds two tests using exactly the technique `knowledgeSearchLowRatedCaveat.test.ts` uses for issue #432 (mock `pool.query` to reject on the one SQL fragment unique to the supplementary lookup, pass everything else through, assert the handler still replies normally).

- **Conflict badge** — two hits both clearing the 0.35 floor, so `relevantIds.length >= 2` genuinely reaches `hasConflictAmongIds` instead of short-circuiting; the `JOIN knowledge b ON a.id < b.id` self-join is rejected. Asserts both hits still render and the caveat is absent: a failed check must not claim entries disagree.
- **Lexical fallback** — one hit deliberately below the floor, which is the only branch reaching `searchKnowledgeLexical`; the `word_similarity(...)` query is rejected. Asserts the handler *resolves* to the normal no-match reply rather than throwing, since a thrown handler is what the MCP layer converts into an `isError` result carrying `error.message`.

Both also assert the raw driver string never appears in the tool result, which is the outbound-exposure half of the concern.

New file rather than an addition to the low-rated one, so the low-rated caveat stays at its default-off value here and `security-floor.json` gains its own alphabetically-placed entry (876 → 878) instead of a shared hunk — the merge-conflict rationale the manifest documents. Regenerated with `npm run test:security:fix`, not hand-counted.

Both are DB-backed and `{ skip }` without `DATABASE_URL`, same as the precedent, so CI's pgvector container is what actually exercises them. Flagging honestly: I have no Postgres in this environment, so they are **verified as skipping cleanly but not yet verified as passing** — CI on this push is their first real run. I'll drive any failure to green.

## CI failure on `90a91a0` — pre-existing flake, not this diff

The `build` job failed on two tests, both `4 !== 2`:

- `repeatMaxTurnsShortcutRouter.test.ts:327`
- `repeatQuestionShortcutRouter.test.ts:418`

Both are the known shared-table interference race those files already document and ship a `retryOnSharedTableInterference` helper for — the assertion counts outbound replies to the shared `super-1` identity across the whole `interactions` table over a sliding 24h window, so it cannot be scoped to its own `conversation_id`, and a parallel test file's insert for that same identity shifts the delta. Issue #675 recorded the `3 !== 2` form of it; this run hit `4 !== 2` (two interfering inserts). The retry helper was exhausted rather than absent.

Unrelated to this diff on both sides: nothing here touches the router, interaction recording, or those tests, and the new test file writes only to `knowledge`/`knowledge_gaps` under a unique per-run id, never `interactions` — so it cannot aggravate that race either. Per `docs/PIPELINE.md` the right move on an unrelated flake is a re-run rather than a fabricated fix; the ci-retry loop's free rerun was already in flight, and this push supersedes it with a fresh run.

If `4 !== 2` keeps recurring, the durable fix is making that budget assertion resilient to concurrent identities rather than retrying — happy to open a separate issue, but it's out of scope here.

---

swampratnz:
Follow-up on the `build` failure for `0348ab9` — short version: the new tests are fine, the flake recurred.

**The two new `SECURITY:` tests passed against real Postgres**, which closes the caveat I flagged when I pushed them (they had only ever been verified as skipping locally):

```
✔ SECURITY: knowledge_search omits the conflict badge and still returns its hits
  when the conflict-check query rejects, never surfacing the DB error (issue #389) (193.3ms)
✔ SECURITY: knowledge_search degrades to a clean no-match instead of an error
  when the below-floor lexical fallback query rejects (issue #362) (48.8ms)
```

The fixture concern I called out — a stray `global` entry outranking the below-floor row and skipping the lexical branch — did not materialise.

**The only failure was `repeatMaxTurnsShortcutRouter.test.ts:327`, `3 !== 2`** — issue #675's exact recorded signature, one interfering insert this time rather than the two that produced `4 !== 2` on the previous commit. `repeatQuestionShortcutRouter` passed on this run. Its `retryOnSharedTableInterference` helper was exhausted, not absent. The ci-retry loop has already taken its free machine rerun (attempt 2, in flight), so I'm letting that settle rather than pushing a speculative fix mid-rerun.

If it fails a third time it stops being worth re-running, and the durable fix is to stop sharing the counted identity: that assertion calls `countRepliesToUser('discord', 'super-1')`, which aggregates `interactions` rows by `meta->>'replyToUserId'` across the whole table for 24h, and `super-1` is the conventional super-admin fixture id used by many test files in parallel. Giving that one test a unique per-run identity (registered via the `SUPER_ADMIN_DISCORD_IDS` env the file already sets, since the runner isolates env per file) makes interference structurally impossible instead of improbable — strictly better than retrying. That's a change to a test this PR doesn't otherwise touch, though, so I'd rather do it as its own PR against #675 than expand an already-approved diff, unless a maintainer would prefer it folded in here.

---

Issue #647, opened 2026-07-20
usage_stats: add optional platform filter to drill into one platform's top-users/cost-by-role — #580's own named growth path, still unbuilt


swampratnz:
## Problem statement

**Who it helps**: the super admin running both a Discord server and a WhatsApp number, who uses `usage_stats` as the primary on-demand cost/volume monitoring tool.

**Evidence, verified against `main`**: issue #580 (closed, built, `theme:admin-insight`) added an always-on `By platform: ...` breakdown line to `usage_stats`'s totals (`src/agent/tools.ts`'s `formatUsageByPlatformLine`, `src/storage/repository.ts`'s `usageStats()` `byPlatform` query). But `topUsers` and `costByRole` — the two most actionable fields in the tool, the ones an admin actually drills into when chasing a cost spike — are still computed as one blended total across every platform, with no way to scope them. #580's own "Growth path (not in this proposal)" section names this exact gap verbatim:

> "an optional `platform` filter param on `usage_stats` itself (mirroring `engagement_stats`'s existing pattern exactly) so a super admin can drill into one platform's top-users/cost-by-role too — natural follow-on once the base breakdown line proves useful, deliberately left out here to keep this one PR small."

The base breakdown line has been live since 2026-07-19 with no follow-up filed. `CHANGELOG.md`'s #580 entry confirms the shipped scope stopped short: "Same `super_admin` gate, no new tool or param." The precedent already exists in this exact codebase — `engagement_stats` (`tools.ts`) already takes an optional `platform: z.enum(['discord', 'whatsapp']).optional()` param and threads it into `engagementStats(args.platform)` — so this isn't new design, it's applying an established, already-reviewed pattern to the one sibling tool that still lacks it.

**Concrete gap this leaves**: after seeing `By platform: discord: 301 in / 290 out, ~$2.40 · whatsapp: 111 in / 108 out, ~$0.72`, a super admin who wants to know *which* users or *which* role is driving WhatsApp's share still has to eyeball the blended `topUsers`/`costByRole` lists and guess — there's no way to ask the tool directly.

`docs/COMMUNITY-CONTEXT.md` has no export generated yet, so this is grounded in direct code/issue-history inspection (VISION evidence tier 3), not community-context data.

## Proposed approach

Mirror `engagement_stats`'s exact param/threading pattern, smallest viable version:

[...]

PR #685, opened 2026-07-25
Deflake reply-budget assertions by scoping them to a run-unique identity

### LINKED ISSUE #684: Reply-budget count interference recurred after #675 — scope the assertions to a run-unique identity instead of retrying
## Problem (measured, post-#675)

#675 deflaked the reply-budget assertions in the two shortcut-router tests with a bounded `retryOnSharedTableInterference`. That shipped and closed as `status:built` on 2026-07-23. **The interference has recurred, exhausting the retry**, observed on PR #683 CI:

- Run 1 (`90a91a0`) — **two** files failed together: `repeatMaxTurnsShortcutRouter.test.ts:327` and `repeatQuestionShortcutRouter.test.ts:418`, both `4 !== 2`.
- Run 2 (`0348ab9`) — `repeatMaxTurnsShortcutRouter.test.ts:327`, `3 !== 2` (#675's exact recorded signature).

Both runs burned the ci-retry rerun. The diff under test was unrelated (a `knowledge_search` guard plus test hygiene), so this is pure interference, not a regression.

## Root cause (unchanged from #675, but not addressed by its fix)

The assertions call `countRepliesToUser('discord', 'super-1')`. That helper counts `interactions` rows by `meta->>'replyToUserId'` for an identity across the **whole table** over a sliding 24h window — it cannot be conversation-scoped, because a daily reply budget is per-user by definition.

`'super-1'` is the conventional super-admin fixture id used by many test files, and the runner executes test files as parallel processes against one Postgres. So any other file recording an outbound reply to `super-1` on Discord lands between the before/after reads. The retry reduces the probability but cannot eliminate it, and under CI contention it is being exhausted.

## Fix

Apply the option #675 itself ranked first but did not take:

> **Prefer scoping over retrying**: if the underlying count query used by the assertion can be filtered by the test's own run-unique `user_id`/`conversation_id`, do that — deterministic, no absorb logic needed.

Give each budget assertion a per-process identity (`BUDGET_USER_ID`) that no other test file can write, registered via the `SUPER_ADMIN_DISCORD_IDS` env each file already sets (the runner isolates env per file), so the identity resolves to the same tier `'super-1'` did and the exercised router path is unchanged. There is then no path by which a concurrent insert can reach the asserted count.

`retryOnSharedTableInterference` is kept in both files as a backstop for their other reads, still `console.warn`-ing per absorbed attempt.

## Acceptance criteria

1. **No `src/` changes** — `git diff --stat` shows only `tests/**`.
2. `tests/repeatMaxTurnsShortcutRouter.test.ts` and `tests/repeatQuestionShortcutRouter.test.ts` budget assertions each count an identity unique to the running process; no other test file can write that identity.
3. Genuine-regression safety preserved: each assertion still pins the **exact** expected count, so an unrecorded shortcut reply still fails.
4. Interference remains observable via `console.warn` where the retry backstop still applies.
5. **`tests/security-floor.json` is NOT modified** — neither assertion is `SECURITY:`-prefixed.
6. `npm run typecheck`, `npm run lint`, `npm test`, and `npm run build` green.

## Relationship to #675

Same defect class and same two assertions, but #675 is closed as completed and its shipped remedy (retry) is the thing being replaced — so this is tracked separately rather than reopening. Scope is deliberately identical to #675's: the shared-table row-count interference class only. The `#422` embedding-similarity flake and per-file Postgres `search_path` isolation remain out of scope, as they were there.

swampratnz:
Closes #684. Follow-up to #675, whose retry-based remedy has been exhausted twice in CI.

## Summary

The reply-budget assertions in both shortcut-router tests count via `countRepliesToUser`, which aggregates outbound replies for an **identity** across the whole `interactions` table over a sliding 24h window. That read cannot be conversation-scoped — a daily budget is per-user by definition — so counting the shared `'super-1'` fixture id absorbed any concurrent insert another test file made for that same identity, and test files run as parallel processes against one Postgres.

#675 addressed this with a bounded retry. It recurred after that shipped, on PR #683's CI:

- `90a91a0` — both files, `4 !== 2`
- `0348ab9` — `repeatMaxTurnsShortcutRouter.test.ts:327`, `3 !== 2` (#675's exact recorded signature)

Both runs exhausted the retry and burned the ci-retry rerun, on a diff that touches none of this.

#675's own acceptance criteria ranked scoping **above** retrying — *"Prefer scoping over retrying … deterministic, no absorb logic needed"* — and this applies that first option. Each budget assertion now counts a per-process `BUDGET_USER_ID` that no other test file can write, registered through the `SUPER_ADMIN_DISCORD_IDS` env each file already sets (the runner isolates env per file), so the identity resolves to the same tier `'super-1'` did and the exercised router path is unchanged.

`retryOnSharedTableInterference` is deliberately kept in both files as a backstop for their other reads, still emitting a `console.warn` per absorbed attempt, and each assertion still pins the exact expected count so a genuine regression fails deterministically. Their doc comments are updated, since they described counting `'super-1'` — no longer true of the budget path.

## Security / privacy impact

None. Test-only: no `src/` changes, so router and shortcut production behaviour is byte-for-byte untouched. No change to auth, tool gating, data-access scope, or stored data.

`tests/security-floor.json` is deliberately **not** modified — neither assertion is `SECURITY:`-prefixed, so no count changes. The only tier-adjacent detail is that the new fixture identity is registered as a super admin, purely so it resolves to the same tier the previous fixture id did; it exists only inside these two test files' env.

## How verified

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm test` — 2419 tests, 596 skipped. 3 failures remain, all pre-existing on `main` and **unrelated to this diff** (two `docsIngest` and one `agentCoreKnowledgeEntryId` test that reach a live DB without `DATABASE_URL`); they are fixed by PR #683 and are not touched here.
- `npm run build` — clean
- Scope confirmed: `git diff --stat` lists only the two test files.

Verified by inspection that no path remains by which another file's concurrent `interactions` insert can reach either asserted count, since the counted identity is unique per process.

The flake itself is probabilistic, so CI passing once is not proof — but the argument here is structural rather than statistical: the counted identity is no longer shared, so there is nothing left to interfere.

---

### DIFF
--- a/tests/repeatMaxTurnsShortcutRouter.test.ts
+++ b/tests/repeatMaxTurnsShortcutRouter.test.ts
@@ -16,7 +16,19 @@ process.env.DISCORD_GUILD_ID ??= '1';
 const hasDb = Boolean(process.env.DATABASE_URL);
 process.env.DATABASE_URL ??= 'postgres://test:test@127.0.0.1:5432/test';
 process.env.WHATSAPP_PROVIDER ??= 'disabled';
-process.env.SUPER_ADMIN_DISCORD_IDS ??= 'super-1,super-2';
+// Run-unique identity used ONLY by the reply-budget assertion below.
+// `countRepliesToUser` aggregates outbound replies for an IDENTITY across the
+// whole `interactions` table over a sliding 24h window — that is what a daily
+// budget means, so the read cannot be conversation-scoped. Counting the shared
+// 'super-1' fixture id therefore absorbs any concurrent insert another test
+// file makes for that same identity, and test files run as parallel processes
+// against ONE Postgres. A per-process id no other file can ever write makes
+// interference structurally impossible rather than merely improbable (issue
+// #675 preferred exactly this scoping over the retry it ultimately shipped).
+// Registered as a super admin so the identity still resolves to the same tier
+// 'super-1' did, keeping the exercised router path unchanged.
+const BUDGET_USER_ID = `budget-super-${process.pid}-${Date.now()}`;
+process.env.SUPER_ADMIN_DISCORD_IDS ??= `super-1,super-2,${BUDGET_USER_ID}`;
 process.env.SUPER_ADMIN_WHATSAPP_NUMBERS ??= 'super-1';
 process.env.REPEAT_MAX_TURNS_SHORTCUT_ENABLED = 'true';
 
@@ -46,13 +58,18 @@ const RUN = `repeatmt-router-${Date.now()}`;
  * interactions table over a sliding 24h window, by design (that's what a
  * daily reply budget means), so it can't be scoped to this test's own
  * conversation_id. The Node test runner executes test FILES in parallel
- * against one shared DB, so another file's concurrent insert for the same
- * 'super-1' discord identity can land between the before/after reads and
- * shift the exact delta (issue #675: "3 !== 2"). Re-running the whole
- * read-seed-read sequence gets a quiet window with overwhelming
- * probability, while a REAL regression fails deterministically on every
- * attempt — so retrying masks nothing (mirrors
- * tests/repository.test.ts's retryOnSharedTableInterference).
+ * against one shared DB, so another file's concurrent insert for the SAME
+ * identity can land between the before/after reads and shift the exact delta
+ * (issue #675: "3 !== 2", which recurred after this retry shipped).
+ *
+ * The primary defence is now `BUDGET_USER_ID`: the budget assertion counts a
+ * per-process identity no other test file can write, so there is no longer a
+ * path by which a concurrent insert reaches the count. This retry is kept as
+ * a backstop for any future interference this file's other reads might hit,
+ * and because a REAL regression still fails deterministically on every
+ * attempt — so retrying masks nothing (mirrors tests/repository.test.ts's
+ * retryOnSharedTableInterference). Every absorbed attempt still emits a
+ * console.warn, so interference is never silent.
  */
 async function retryOnSharedTableInterference(attempts: number, run: () => Promise<void>): Promise<void> {
   for (let attempt = 1; ; attempt++) {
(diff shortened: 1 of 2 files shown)


swampratnz:
`build` failed on `7c741ca`, but **this PR's own fix is validated** — the failure is a different assertion.

## The fix works

`pass 2418, fail 1`. Both target assertions passed on their first CI run:

- `repeatMaxTurnsShortcutRouter.test.ts` reply-budget — **passed** (failed `3 !== 2` on #683's `0348ab9`, `4 !== 2` on `90a91a0`)
- `repeatQuestionShortcutRouter.test.ts` reply-budget — **passed** (failed `4 !== 2` on `90a91a0`)

No `console.warn` from `retryOnSharedTableInterference` on either, so they didn't merely survive on a retry — the counts were clean first time, which is what identity-scoping predicts.

## The one failure is a different mechanism, not a missed call site

`tests/repository.test.ts:3343` — *"discord-scoped topUsers includes the seeded discord user"* (`usageStats`, issue #647). It exhausted that file's own `retryOnSharedTableInterference`.

Same broad family (shared-table interference), but **not the same defect, and this PR's technique cannot fix it**:

```sql
GROUP BY user_id ORDER BY n DESC LIMIT 5
```

`topUsers` is a **top-5 leaderboard**, not a count delta. The test seeds one message for a run-unique `discordUser` and asserts it appears. It already has a unique identity — the problem is that a single message competes for a top-5 slot against every other test file's concurrent Discord traffic, and gets evicted. Scoping the identity is exactly what's already true there and doesn't help; the fix would be something else (seed enough messages to guarantee a slot, or assert against a predicate that isn't rank-limited).

That is outside #684's stated scope, which bounds this to the two shortcut-router files — the same boundary #675 set. I'd rather not widen an approved diff to chase it; happy to file it as its own issue in the same series, which looks like the third distinct instance of this class after #675 and #684.

ci-retry has taken its free rerun (attempt 2, in flight). Letting that settle rather than pushing anything, since nothing in this diff is implicated.

---

PR #688, opened 2026-07-25
deps: clear 4 of 5 security advisories (patch bumps + sharp override)

swampratnz:
Triage of the Dependabot alerts on `main`, then the fixes that were worth taking.

**`npm audit` goes from 7 vulnerabilities (3 high, 4 moderate) to 3 moderate** — and those 3 are a single advisory counted across its dependency chain.

Up front: **none of the four distinct advisories was reachable in this deployment.** This is hygiene, not incident response. Three of them cost nothing to close, so I closed them.

## Taken

**`brace-expansion` 5.0.7 → 5.0.8** — [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg), high (CVSS 7.5). Lockfile only. Highest CVSS of the set and the lowest real risk: reached solely via `eslint → minimatch`, `npm ls --omit=dev` is empty so it never shipped, and its only input is this repo's own lint globs.

**`protobufjs` 7.6.4 → 7.6.5** — [GHSA-j3f2-48v5-ccww](https://github.com/advisories/GHSA-j3f2-48v5-ccww), moderate. Lockfile only. The DoS is in the `.proto` option **parser**; Baileys imports `protobufjs/minimal.js` against precompiled static codecs, and the `minimal` build excludes the parser entirely.

**`sharp` override `^0.35.3`** — [GHSA-f88m-g3jw-g9cj](https://github.com/advisories/GHSA-f88m-g3jw-g9cj), high. npm reports *"no fix available"* because `@huggingface/transformers` pins `0.34.5`. But Baileys already ran `sharp@0.35.2` in the same tree, so an override was demonstrably viable rather than speculative — and it dedupes the two copies to one (`0.35.3`, libvips 8.18.3). Exploiting the libvips CVEs requires decoding an **attacker-supplied image**, and no image reaches sharp here: transformers is used only for `feature-extraction` (text embeddings) and `automatic-speech-recognition` (audio), and image generation shells out to a sandboxed `grok` subprocess.

## Deliberately not taken

**`@hono/node-server`** — [GHSA-frvp-7c67-39w9](https://github.com/advisories/GHSA-frvp-7c67-39w9), moderate. Windows-only path traversal in `serve-static`, unreachable twice over: deployment is Ubuntu/systemd per `docs/DEPLOYMENT.md`, and the MCP server is **in-process** (`createSdkMcpServer`), so no hono server is ever started and `serve-static` is never mounted.

npm's remedy is `@anthropic-ai/claude-agent-sdk@0.2.85` — a **downgrade across a major version** from the pinned `^0.3.201`. That's a real regression to dodge a vulnerability that cannot fire here, so the alert stays open by choice. Worth a Dependabot ignore rule if the noise is annoying.

## Security / privacy impact

No source changes at all — `package.json` (one `overrides` entry), `package-lock.json`, and `CHANGELOG.md` only. No change to auth, tier derivation, tool gating, data-access scope, outbound filtering, or stored data. `tests/security-floor.json` untouched (881 across 111 files, unchanged — no `SECURITY:` test added or removed).

The `sharp` override is the only change with runtime reach, and it moves a native module **forward** to the patched libvips.

## How verified

Standard gate, all green:

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm test` — 2425 tests, **0 fail** (1825 pass, 600 skip)
- `npm run test:security` — 881 across 111 files, manifest matches
- `npm run build` — clean

Because the `sharp` override replaces a native module that `transformers` pinned, the test suite alone isn't sufficient evidence — nothing in it exercises the embedding model. Two extra checks:

- Loaded the **real embedding pipeline** end to end (`Xenova/all-MiniLM-L6-v2`): returned a 384-dim unit vector, norm `1.0000`. This is the check that would catch transformers failing to import or initialise against the newer sharp.
- Exercised **sharp's native binding** directly: `sharp.versions` reports `0.35.3` / libvips `8.18.3`, and an encode round-trip succeeded.

Not verified: Baileys' own thumbnail path, which uses sharp but has no test coverage here. It was already on `0.35.2`, so this is a patch bump for that consumer.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,21 @@ for anything after ~noon NZST/NZDT). Get today's date with
   neither throw nor reveal the configured token's length. Low severity — a
   one-off setup handshake where network jitter dominates the signal — but it
   removes the last early-exit secret comparison on that path.
+- **Dependency advisories cleared: 7 down to 3.** `brace-expansion` (5.0.8)
+  and `protobufjs` (7.6.5) are lockfile-only patch bumps, and a new `sharp`
+  override (`^0.35.3`) resolves the libvips CVEs npm reported as having no
+  fix, because `@huggingface/transformers` pinned `0.34.5`; that also dedupes
+  the two `sharp` copies in the tree to one. None of the three was reachable
+  in this deployment — `brace-expansion` is dev-only via eslint, the
+  `protobufjs` parser is absent from the `protobufjs/minimal` build Baileys
+  uses, and no image ever reaches `sharp` (transformers is used only for text
+  embeddings and speech, and `/imagine` shells out to a sandboxed subprocess)
+  — so this is hygiene, not incident response. The remaining advisory
+  (`@hono/node-server`, Windows-only `serve-static` path traversal) is
+  deliberately NOT taken: it is unreachable twice over (Ubuntu deployment; the
+  MCP server is in-process via `createSdkMcpServer`, so no HTTP server is ever
+  started), and npm's suggested remedy downgrades
+  `@anthropic-ai/claude-agent-sdk` across a major version.
 
 ## 2026-07-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,11 +70,16 @@ for anything after ~noon NZST/NZDT). Get today's date with
   neither throw nor reveal the configured token's length. Low severity — a
   one-off setup handshake where network jitter dominates the signal — but it
   removes the last early-exit secret comparison on that path.
-- **Dependency advisories cleared: 7 down to 3.** `brace-expansion` (5.0.8)
-  and `protobufjs` (7.6.5) are lockfile-only patch bumps, and a new `sharp`
-  override (`^0.35.3`) resolves the libvips CVEs npm reported as having no
-  fix, because `@huggingface/transformers` pinned `0.34.5`; that also dedupes
-  the two `sharp` copies in the tree to one. None of the three was reachable
+- **Dependency advisories cleared: 7 down to 3.** Three new `overrides`
+  entries — `brace-expansion` (`^5.0.8`), `protobufjs` (`^7.6.5`) and `sharp`
+  (`^0.35.3`). The `sharp` one resolves the libvips CVEs npm reported as
+  having no fix, because `@huggingface/transformers` pinned `0.34.5`; it also
+  dedupes the two `sharp` copies in the tree to one. Overrides rather than an
+  `npm audit fix` lockfile regen, deliberately: the regen silently carried
+  `@anthropic-ai/claude-agent-sdk` forward 0.3.201 → 0.3.220 inside its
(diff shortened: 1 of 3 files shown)


swampratnz:
Good catch, and it was right — verified both against the lockfile diff before acting. Fixed in `dfe1c31`.

## The drift was real

`npm audit fix` carried two packages forward inside their existing semver ranges, neither mentioned anywhere in the diff's stated intent:

- `@anthropic-ai/claude-agent-sdk` **0.3.201 → 0.3.220**, plus all eight per-platform optional deps
- `@hono/node-server` **1.19.14 → 1.19.15**

The SDK one is the serious half, for exactly the reason given: it's what enforces `tools: []` per turn, the tier-derived tool surface, and the confirm-before-destructive flow. It also made the PR internally inconsistent — the same change declines to let npm *downgrade* that package for the hono advisory, so accepting an incidental 19-version *upgrade* of it was not defensible.

## Removed rather than disclosed

Disclosure was the minimum ask, but the better answer is for the diff to do only what it claims. The lockfile is reset to `main`'s and the three fixes are now `overrides` entries instead — narrower (npm re-resolves nothing else) and self-documenting in `package.json`, matching the existing `adm-zip`/`undici` precedent.

Net lockfile change vs `main` is now exactly:

| Package | Change | Why |
|---|---|---|
| `brace-expansion` | 5.0.7 → 5.0.8 | advisory fix |
| `protobufjs` | 7.6.4 → 7.6.5 | advisory fix |
| `sharp` | 0.35.2 → 0.35.3, nested 0.34.5 copy collapsed | advisory fix + dedup |
| `@img/*` | sharp's per-platform packages | consequence of the above |
| `@emnapi/runtime` | 1.11.1 → 1.11.2 | transitive of `@img/sharp-wasm32` |

`claude-agent-sdk` and `@hono/node-server` are untouched — **zero references** in the lockfile diff (`git diff origin/main -- package-lock.json | grep -cE "claude-agent-sdk|hono/node-server"` → `0`). The `@humanfs/*` and `@humanwhocodes/*` lines that appear in the diff are identical versions merely relocated by the regen, not version changes.

Since the bump is gone, the suggested SDK-changelog review is moot for this PR. If a deliberate 0.3.220 upgrade is wanted, it should be its own change where the gating-behaviour diff can be the actual subject of review rather than a footnote.

## Re-verified after the rebuild

Typecheck, lint, format, build clean; `npm test` 2425 tests / **0 fail**; `test:security` 881 across 111 files; `npm audit` still 3 moderate (the single hono advisory across its chain); `@anthropic-ai/claude-agent-sdk` confirmed installed at **0.3.201**.

Both extra checks re-run against the rebuilt tree, since the override still moves a native module: the live embedding pipeline returned a 384-dim unit vector (norm `1.0000`), and sharp's native binding round-tripped an encode on `0.35.3` / libvips `8.18.3`.

One correction to the review for the record: it reported the deduped sharp as "libvips `1.3.2`" — `sharp.versions.vips` reports `8.18.3`. Doesn't affect the conclusion.

---

PR #690, opened 2026-07-25
deps: bump @anthropic-ai/claude-agent-sdk 0.3.201 → 0.3.220

swampratnz:
The deliberate version of the bump #688 backed out. This is the package that enforces the repo's tool-gating invariants, so **the behavioural review is the substance here, not the version number.**

## How it was reviewed

The package ships no changelog, so I diffed the shipped type declarations between the two versions. The four levers `agent/core.ts`'s `buildQueryOptions` depends on are **byte-identical** — same names, same types, same doc comments:

| Lever | Used for |
|---|---|
| `tools` | built-in allowlist — `[]`, or `['WebSearch']` for admin+ |
| `allowedTools` | tier-derived auto-approve list |
| `disallowedTools` | `Task`, `WebFetch`, and `WebSearch` below admin |
| `permissionMode` / `PermissionMode` | `'default'` |

`tool()` and `CreateSdkMcpServerOptions` are unchanged as well, so the in-process MCP server in `agent/tools.ts` is unaffected.

## One real behavioural change

The SDK MCP **tool-call timeout moved**:

- 0.3.201 — *"If your SDK MCP calls will run longer than 60s, override `CLAUDE_CODE_STREAM_CLOSE_TIMEOUT`"*
- 0.3.220 — *"Tool calls are bounded by the MCP tool-call timeout — the `MCP_TOOL_TIMEOUT` env var (ms), effectively unbounded by default"*

So the implicit ~60s ceiling on a tool call is gone, and this repo sets neither env var.

I judged this acceptable rather than fixing it here: every community tool is a Postgres query, and `storage/db.ts` already bounds those independently via `statement_timeout`, `query_timeout` and `connectionTimeoutMillis` (all `config.db.*`). A hung tool call therefore still terminates at the DB layer rather than relying on the SDK's timeout. Worth revisiting if a non-DB tool is ever added — happy to set `MCP_TOOL_TIMEOUT` explicitly if you'd prefer a belt-and-braces bound now.

## Everything else is additive and inert here

New hook input types, new telemetry fields, a `matchedAskRule` field on the `canUseTool` callback (unused — `core.ts` gates via `allowedTools`, not `canUseTool`), and **three new built-in tools**: `RefreshMcpTools`, `SendFeedback`, `ProposeSkills`.

Those new built-ins are worth a sentence, because a growing built-in surface is exactly the thing that could quietly widen a turn's capability. They're structurally unreachable: `tools` is an explicit **allowlist**, not a denylist, so anything the SDK adds is excluded by default. That's precisely what the existing `SECURITY: members/guests get NO built-in tools; admin+ get exactly WebSearch` test pins — and it passes unchanged against the larger surface, so the allowlist model is verified rather than merely asserted.

## Security / privacy impact

No source changes — `package.json` and `package-lock.json` only. No change to tier derivation, tool gating, data-access scope, outbound filtering, confirm flow, or stored data. `tests/security-floor.json` untouched (881 across 111 files).

The security-relevant surface *is* this dependency, which is why the review above is the deliverable. Net finding: the gating contract is unchanged, one timeout default relaxed with an existing independent backstop.

## How verified

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm test` — 2425 tests, **0 fail** (1825 pass, 600 skip)
- `npm run test:security` — 881 across 111 files, manifest matches
- `npm run build` — clean

Specifically re-ran the 14 tool-gating assertions in `tests/agentOptions*.test.ts` against the new SDK — tier surfaces, `WebFetch`/`Task` denial for every tier, empty `settingSources`, and the `allowedTools`-vs-`rbac.ts` drift check. All pass.

Scope check: `git diff --stat` shows only `package.json` and `package-lock.json`; the lockfile moves the SDK and its eight per-platform optional deps and nothing else — no repeat of #688's incidental drift.

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "community-agent",
       "version": "0.1.0",
       "dependencies": {
-        "@anthropic-ai/claude-agent-sdk": "^0.3.201",
+        "@anthropic-ai/claude-agent-sdk": "^0.3.220",
         "@hapi/boom": "^10.0.1",
         "@huggingface/transformers": "^4.2.0",
         "@whiskeysockets/baileys": "6.7.23",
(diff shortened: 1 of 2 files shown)


swampratnz:
Fair flag on not being able to re-verify the `.d.ts` comparison — that claim shouldn't rest on my word for the package that enforces tool gating. Here is the check in a form anyone can reproduce from published tarballs, no repo checkout or `node_modules` needed:

```bash
mkdir /tmp/sdkverify && cd /tmp/sdkverify
npm pack @anthropic-ai/claude-agent-sdk@0.3.201
npm pack @anthropic-ai/claude-agent-sdk@0.3.220
mkdir a b
tar xzf anthropic-ai-claude-agent-sdk-0.3.201.tgz -C a
tar xzf anthropic-ai-claude-agent-sdk-0.3.220.tgz -C b
diff -u a/package/sdk.d.ts b/package/sdk.d.ts
```

## Stronger result than the PR body claimed

Rather than checking the four levers in isolation, I extracted the **entire `Options` type** (the whole 732-line block, brace-matched) from both versions and diffed it. It is identical except for **four documentation-URL updates**:

```diff
-     * @see https://docs.anthropic.com/en/api/beta-headers
+     * @see https://platform.claude.com/docs/en/api/beta-headers
-     * @see https://docs.anthropic.com/en/docs/build-with-claude/adaptive-thinking
+     * @see https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking
-     * @see https://docs.anthropic.com/en/docs/build-with-claude/effort
+     * @see https://platform.claude.com/docs/en/build-with-claude/effort
-     * @see https://docs.anthropic.com/en/docs/claude-code/settings#sandbox-settings
+     * @see https://code.claude.com/docs/en/settings#sandbox-settings
```

Zero semantic changes across the entire options surface — not just `tools` / `allowedTools` / `disallowedTools` / `permissionMode`, but every option the SDK accepts. That is a broader guarantee than "the four levers match", and it's mechanically checkable with the commands above.

To reproduce just that block:

```bash
python3 - <<'PY'
import re
def opts(p):
    s=open(p).read(); m=re.search(r'export declare type Options = \{', s)
    i=m.end(); d=1
    while d:
        d += (s[i]=='{') - (s[i]=='}'); i+=1
    return s[m.start():i]
open('a.opts','w').write(opts('a/package/sdk.d.ts'))
open('b.opts','w').write(opts('b/package/sdk.d.ts'))
PY
diff -u a.opts b.opts
```

## Caveat I want to keep honest

This verifies the **declared type surface**, not the compiled implementation — `sdk.mjs` is ~915 KB of bundled code and did change between versions. So this establishes that the gating *contract* is unchanged and that the repo cannot be silently passing options with different meanings; it does not prove the runtime enforces them identically. What covers that gap is the repo's own behavioural tests, which assert enforcement rather than shape — the 14 gating assertions in `tests/agentOptions*.test.ts` (tier surfaces, `WebFetch`/`Task` denial for every tier, empty `settingSources`, `allowedTools`-vs-`rbac.ts` drift) all pass against 0.3.220.

The timeout claim comes from the same diff, in `createSdkMcpServer`'s doc comment — visible in the `diff -u` of the full `sdk.d.ts` above.

## CI

`lint`, `security-invariants`, `review`, and `discover` are now green. `build` is still running — that's the one exercising the new SDK against the pgvector container, so it's worth waiting on before merge, as you noted.

---

Issue #613, opened 2026-07-20
Docs-ingest: aggregate per-page fetch-failure warnings into one run-end summary


swampratnz:
## Problem statement

**Who it helps**: the operator/super-admin, who reads the bot's logs to gauge health. Theme: **Reliability & ops** (`docs/VISION.md`: "resilience, observability, graceful degradation").

**Evidence** (from community-feedback issue #611, filed by the repo owner from an actual run): the weekly docs-ingest job logged **157 of 586 page fetches failing with HTTP 404 (~27%)**, one `logger.warn({ err, url }, 'Docs ingest: page fetch failed')` line per page (`src/context/docsIngest.ts:283`), every one of them under `https://platform.claude.com/docs/en/api/terraform/beta/...`. Root cause is upstream — Anthropic's `llms.txt` index lists a batch of `terraform/beta/*` pages that don't actually exist — not a bug in this codebase, and the run still completed correctly (`fetched: 429, created: 159, updated: 899, removed: 18`; `runDocsIngest`'s own doc comment: *"Never throws on a single page's failure — it's counted and the run continues"*). So this is pure log-noise/signal-to-noise degradation, not a functional break: a genuine new failure class (e.g. a real outage, a broken selector) would be buried in 157 near-identical lines.

`result.failed` (`DocsIngestResult`) already aggregates the *count* — nothing currently aggregates what gets *logged*.

## Proposed approach

In `runDocsIngest` (`src/context/docsIngest.ts`), stop calling `logger.warn` per failed page fetch inside the `worker` closure. Instead:

1. Collect failed-fetch URLs in a local array as the pool runs (still increment `result.failed` exactly as today — no `DocsIngestResult` shape change).
2. After `runPool` completes, if any page fetches failed, emit **one** `logger.warn` summary line: total failed / total attempted, a capped sample of the first ~5 failing URLs, and a by-path-prefix rollup (group by `docPathOf(url)`'s first segment, e.g. `147× under api/terraform/beta`).
3. Keep full per-page detail available at `debug` level (one `logger.debug({ err, url }, ...)` per failure, unchanged shape) so nothing is lost for someone who turns debug logging on — only the default `warn`-level output changes.
4. `result.failed`, `result.fetched`, and every other counted field are untouched — this is a pure logging change.

[...]

PR #691, opened 2026-07-25
Skip docs-ingest URLs that have 404'd for N consecutive runs (#611)

### LINKED ISSUE #611: Docs-ingest: aggregate upstream 404s into one summary warning (not one per page)
## Observed (2026-07-18 ingest run)

The weekly docs-ingest run logged **157 of 586 pages failing HTTP 404 (~27%)**, one warn-level line each — every failure under `https://platform.claude.com/docs/en/api/terraform/beta/...`. The run still completed correctly (`fetched: 429, created: 159, updated: 899, removed: 18`), so this is log-noise + missing-coverage, not a functional break.

Root cause is **upstream**: Anthropic's `llms.txt` index lists a batch of `terraform/beta/*` API pages that don't exist. Not a bug in Dave.

## Proposed

In `src/context/docsIngest.ts`, don't emit a warn per failed page. **Aggregate** fetch failures and emit a single summary at run end, e.g.:
`Docs ingest: N/M pages failed to fetch` with a small capped sample of URLs (first ~5) and a by-prefix rollup (e.g. `147× under /api/terraform/beta/`). Keep the existing per-page detail available only at debug level.

Optionally: track a rolling "consistently-404 for K runs" set so persistently-dead index entries can be reported once and skipped, rather than re-fetched-and-404'd every week.

## Acceptance criteria
- A run with F failed fetches emits **one** warn-level summary line (count + capped sample + prefix rollup), not F lines.
- The run still completes and ingests all reachable pages (behaviour otherwise unchanged); `indexFetchFailed` semantics preserved.
- Unit test: given a fetcher that 404s a subset, exactly one aggregated warning is emitted and the ingest result counts are correct.

swampratnz:
Closes #611

## Summary

Closes the second, deferred half of #611. The first half — batching the per-page 404 warnings into one run-end summary — shipped as #613, which explicitly left this piece as a growth path because it needs persisted state:

> "the 'consistently-404 for K runs' rolling tracker from #611, to stop re-attempting URLs the index has listed as dead for many consecutive runs — a separate, larger proposal given it needs new persisted state."

Anthropic's `llms.txt` habitually lists pages that don't exist (the run in #611: **157 of 586 returning 404**, all under `api/terraform/beta/*`). #613 made that quiet in the log; the weekly job still re-requested every dead page, every run.

A URL that fails to fetch on `DOCS_INGEST_DEAD_URL_RUNS` consecutive runs (default **3** — ~3 weeks at the weekly cadence) is now reported **once** and then skipped. The skip is **self-healing, not permanent**: each skipped URL is re-probed once every `DOCS_INGEST_DEAD_URL_RECHECK_DAYS` (default 30), and a successful fetch deletes its streak row so it rejoins the normal fetch set with no operator action. `DOCS_INGEST_DEAD_URL_RUNS=0` is a complete off-switch — no streak read, no streak write, no reporting.

### Changes

- **`schema.sql`** — new `docs_ingest_url_failures` table. A row exists only while a URL is *currently* failing (a success deletes it), so it stays bounded by the size of the dead tranche rather than by history, and needs no retention purge. Holds URLs only — no user identifier, so `forget_me`/`purge_user_data` have nothing to touch.
- **`repository.ts`** — `list`/`record`/`clear`/`markReported` helpers, returning raw streak state so the *policy* lives in `docsIngest.ts` (no config coupling in the repo layer). The record upsert de-duplicates its input, because `ON CONFLICT DO UPDATE` errors outright if a key repeats in one statement.
- **`docsIngest.ts`** — pure `partitionDeadUrls()` (state + clock in, decision out) so the skip rule is unit-testable with no DB or network; the store and clock are injectable. New `deadSkipped` result field.
- **`backgroundJobs.ts`** — the stage-2 *"all page fetches failed"* alarm now measures against pages **attempted** (`pages - deadSkipped`), not pages *listed*. Without this, a run whose every remaining page is dead-skipped has `fetched === 0` and would DM super admins a **false total-failure alarm**.

## Security / privacy impact

No new tool, RBAC surface, user data, or external input — this only changes *which already-listed, first-party URLs get requested* on a background job that reads one fixed official index. It cannot widen the fetch set, and it **cannot lose knowledge**: pruning still keys off the index, never off fetch success (the existing fail-safe #611's own report relies on). The new table stores URLs and counters only.

No `SECURITY:`-prefixed test is added and `tests/security-floor.json` is unchanged (888) — same reasoning #613 recorded for the sibling change: there is no tool or data-access surface here for one to pin.

Fail-safe throughout: an unreadable streak store falls back to **fetching everything** (never to skipping blindly), and a bookkeeping write failure is logged without failing a run whose ingest work succeeded.

## How verified

⚠️ **Local caveat:** this sandbox has no Postgres (and no pgvector or Docker daemon to start one), so **all 613 DB-gated tests skip locally** — including the dead-skip regression test. CI is the real check for those; the first CI run duly caught a defect in that test (see below).

- `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run build` — clean.
- `npm run test:security` — **888/888**, manifest unchanged.
- `npm test` — **2449 tests, 0 failures**, *of which 613 DB-gated skipped locally*.

New coverage:
- **Pure policy** (`partitionDeadUrls`, runs everywhere): threshold boundary (`>=`, not `>`), in-cooldown skip vs. due-for-re-probe, `deadRuns=0` disables.
- **Run level** (DB-free — stubbed store + all-fetches-fail): a dead URL costs no request and is counted in `deadSkipped`; a threshold crossing is reported exactly once and stamped; an already-reported URL that re-probes and fails again is **not** re-reported; a config-lowered crossing still reports once before going quiet; a store read failure degrades to fetching everything; `=0` performs no store calls at all.
- **DB-backed repository**: streaks are *consecutive* (a success resets to 1, not cumulative), the upsert bumps rather than duplicating, re-marking preserves the original report time, empty-list calls are no-ops.
- **DB-backed regression guard**: `defaultDocsIngestRun` does **not** throw when every listed page was dead-skipped.

## Review fixes

- **Config-lowered crossing went unreported** (`4069ae9`): the newly-dead report was computed only from URLs that failed *this run*, but lowering `DOCS_INGEST_DEAD_URL_RUNS` onto an existing streak makes a URL dead *before* it is ever re-attempted — so it dropped out of the fetch set silently. Now reported on the crossing however it happened.
- **Dead-skip regression test failed on real Postgres** (`4e6eba8`): it wrote its row at real `now()` then mocked `Date` to `FAR_FUTURE_MS` (+5 years, this suite's convention for the *outer* freshness guard), so the row read as ancient, fell due for re-probe, and was fetched instead of skipped. Product behaviour was correct — a 5-year-old failure *should* be re-probed — so the fix anchors the row to the same mocked instant. Reviewing that also surfaced a second hazard: pre-existing docs-ingest tests call `runDocsIngest` without stub deps and would read/write the real streak table against the shared CI database, so `=0` became a complete off-switch and both affected files pin it off at module scope.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -314,6 +314,16 @@ DOCS_INGEST_CONCURRENCY=5
 # Set empty to ingest the whole corpus. Tune to taste (e.g. drop api/go but keep
 # api/python by removing it from this list).
 DOCS_INGEST_EXCLUDE_PATHS=api/go,api/csharp,api/java,api/python,api/typescript,api/ruby,api/php,api/cli,api/compliance
+# Dead-URL skipping: the upstream index habitually lists pages that don't exist
+# (one observed run: 157/586 404ing, all under api/terraform/beta/*). After this
+# many CONSECUTIVE runs of failing to fetch, a page URL is reported ONCE and
+# then skipped instead of being re-fetched every run. Set 0 to disable skipping
+# (fetch every listed page every run, as before).
+DOCS_INGEST_DEAD_URL_RUNS=3
+# How long a skipped URL stays skipped before one re-probe. This makes the skip
+# self-healing: if upstream restores the page the re-probe succeeds and it
+# rejoins the normal fetch set automatically.
+DOCS_INGEST_DEAD_URL_RECHECK_DAYS=30
 # Knowledge link-rot check: a ~weekly job HEAD-checks every knowledge entry's
 # sourceUrl and flags dead citations for admin review (list_knowledge
 # sourceUnreachable filter). Off by default. sourceUrl is admin-authored only
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,20 @@ for anything after ~noon NZST/NZDT). Get today's date with
   is unset.
 
 ### Fixed
+- **Docs ingest stops re-fetching pages the upstream index lists but doesn't
+  have** (#611): Anthropic's `llms.txt` habitually lists a tranche of dead
+  pages — one observed run had 157 of 586 returning 404, all under
+  `api/terraform/beta/*` — and every weekly run re-requested all of them. A
+  page that fails to fetch on `DOCS_INGEST_DEAD_URL_RUNS` consecutive runs
+  (default 3, so ~3 weeks at the weekly cadence) is now reported **once** and
+  then skipped, instead of costing a request and a log line every week. The
+  skip is self-healing rather than permanent: each skipped page is re-probed
+  once every `DOCS_INGEST_DEAD_URL_RECHECK_DAYS` (default 30), so if Anthropic
+  restores the page it silently rejoins the normal fetch set with no operator
+  action. Set `DOCS_INGEST_DEAD_URL_RUNS=0` to keep fetching everything every
+  run. This completes the second half of #611 — the first half (batching the
+  404 log spam into one summary line) shipped in #613. No knowledge can be lost
+  by a skip: pruning still keys off the index, never off fetch success.
 - **`knowledge_search` no longer throws away a good answer when a secondary
   lookup fails.** The advisory conflict-badge check and the below-floor
   lexical fallback were the only two unguarded database calls in the handler,
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -477,6 +477,15 @@ A normal user tries to get the agent to moderate, announce, or reveal secrets.
   off the **index**, not fetch success — a `'docs'` chunk is removed only when
   its page is no longer listed in the index at all, so a page that transiently
   404s/times out (the index habitually lists some dead URLs) is never deleted.
+  That same dead-URL habit is also bounded on the *fetch* side (issue #611): a
+  URL that fails `DOCS_INGEST_DEAD_URL_RUNS` consecutive runs (default 3, ~3
+  weeks at the weekly cadence; `0` disables) is reported once and then skipped
+  rather than re-fetched every run, with one re-probe every
+  `DOCS_INGEST_DEAD_URL_RECHECK_DAYS` so an upstream fix self-heals with no
+  operator action. This changes only which *already-listed, first-party* URLs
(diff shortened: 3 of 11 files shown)


swampratnz:
Confirmed — you were right, and CI agreed. The `build` job on `54be5f0` failed with exactly the error you predicted:

```
AssertionError [ERR_ASSERTION]: Got unwanted rejection.
Actual message: "Docs ingest: all 1 attempted page fetches failed"
```

Your mechanism was correct in every detail: the mocked `FAR_FUTURE_MS` clock (+5 years) made the row's real-`now()` `last_failed_at` read as ancient, so `dueForRecheck` was trivially true, the URL went back into `toFetch`, and the deliberate `'unreachable'` throw was counted as an ordinary failure — producing the very false alarm the test exists to disprove.

To be clear about the "2447 tests, 0 failures" line in the PR body: that run was real, but this sandbox has no Postgres (and no pgvector or Docker daemon to start one), so all 613 DB-gated tests — including this one — **skipped** locally. The claim was true but not evidence for this test. I've corrected the PR body's verification section to say so.

Fixed in `4e6eba8` using your second option, since the product behaviour was never wrong (a 5-year-old failure genuinely *should* be re-probed) — only the test's setup was. The row is now anchored to the same instant the clock is mocked to, so it sits inside the cooldown.

Reviewing the DB paths for this also turned up a second hazard nothing had flagged: every docs-ingest test written before this feature calls `runDocsIngest` **without** stub deps, so with the feature on they read/write the real streak table against the shared CI database — and a URL failed by enough of them could start being *skipped*, silently changing those tests' expected counts. Two changes for that:

- `DOCS_INGEST_DEAD_URL_RUNS=0` is now a **complete** off-switch — no streak read, no streak write, no reporting — so opting out is byte-identical to pre-feature behaviour with no extra queries (what `.env.example` already promised). Asserted directly.
- Both affected test files pin the feature off at module scope (`config` is per-file, since node:test runs each file in its own process); the dead-URL tests opt themselves back in.

---

Issue #633, opened 2026-07-20
Member-contributed knowledge tips: a suggest_knowledge member tool feeding the existing admin-reviewed candidate queue


swampratnz:
_VISION theme: **Community flywheel** (new, 2026-07-21) — filed under `theme:knowledge` until a dedicated label exists._

## Problem statement

**Who it helps:** every member, twice over — the member who just solved something and wants to save the next person the pain, and every later member whose `knowledge_search` finds the tip.

**Evidence, verified against `main`:** the knowledge base has exactly two write paths today: admin `save_knowledge` (ADMIN_TOOLS, `src/auth/rbac.ts`) and the offline context builder's machine-drafted candidates (`src/context/builder.ts`, issue #102) which flow into `knowledge_candidates` for admin review via `list/accept/decline_knowledge_candidate`. A member has **no path** to contribute durable knowledge: `suggest_improvement` (MEMBER_TOOLS) is explicitly a bot-feature queue, not knowledge — its own tool description and `docs/ARCHITECTURE.md` say so. So when a member posts "for anyone else hitting this: Bedrock cross-region inference needs X", the only way it becomes durable, searchable knowledge is if an admin happens to notice and re-types it via `save_knowledge`, or the context builder's clustering happens to surface it days later (which requires `CONTEXT_CANDIDATES_ENABLED` and ≥`minDistinctUsers` distinct people hitting the theme — a single member's hard-won tip never clears that floor by design). The new VISION mission clause ("contribute to each other") and the flywheel theme name this exact gap; the north-star metrics section lists "member contributions (accepted candidate entries)" as a signal currently pinned at zero.

## Proposed approach

One new MEMBER_TOOLS entry, `suggest_knowledge(title, content)`:

1. **Storage — reuse, don't build:** insert into the existing `knowledge_candidates` table (its `digest_id` is already nullable — `BIGINT REFERENCES context_digests(id) ON DELETE SET NULL`) with `digest_id NULL`, plus two new nullable columns for provenance: `source_platform TEXT` and `source_user_id TEXT` (NULL for existing machine rows — zero behaviour change). Topic = title; run the existing `candidateTopicAlreadyReviewed` + `knowledgeCoversTopic` dedup guards before insert, exactly as `runContextBuilder` does, and tell the member when their tip is already covered (with the covering entry's title) instead of silently queuing a duplicate.
[...]

Issue #667, opened 2026-07-22
Build worker: incremental branch-only pushes (no early PR), so token expiry can't strand finished work


swampratnz:
## Problem

Same measured problem as #663 (rejected as specified, re-proposed here per its own guidance): ~6 approved builds finished green and then lost everything to App-installation-token expiry (~1h lifetime vs the deliberately-generous 120-min build wall-clock) at the final `git push` — full rebuild required on re-queue.

## Proposal — the reviewer's own "materially simpler" alternative, adopted verbatim

**Push the branch incrementally, but do NOT open the PR until the end** (the publish step stays exactly where it is):

1. After the first meaningful commit, and after each subsequent commit (or at minimum right after the gate passes), the worker runs `git push origin HEAD` to its issue branch.
2. The PR is still opened only at the very end, exactly as today.
3. On a publish-step 401, the escalation comment states the branch name and last pushed commit, so a re-queued build (or a human) resumes from the pushed branch instead of rebuilding from scratch.

Why this variant survives the #663 review:
- **Verify-step contract intact**: `gh pr list ... "Closes #N"` still sees no PR for a dead build → run concludes `failure` → `pipeline-build-retry.yml` still re-runs it.
- **Groundskeeper contract intact**: a dead build's issue still has no open PR closing it → still escalated loudly after 3h. No silent-zombie-draft mode exists.
- **Auto-merge unaffected**: nothing changes about when/how PRs appear.
- **No CI amplification**: branch pushes without a PR trigger no `pull_request` workflows; `on: push` is main-only in this repo, so intermediate pushes are free.
- **Single-workflow scope**: only `pipeline-build.yml` changes (prompt + allowed tools already permit the exact `git push origin HEAD` form the other loops use).

## Security / posture

- No new credentials, tokens, or permissions; the same push grant the worker already uses at publish time, used earlier and more often.
- Branch content is exactly what would have been pushed at the end anyway; nothing is merged, reviewed, or executed from it until the normal publish + review path runs.
- Stale branches from dead builds are already handled by the existing branch-hygiene practices (merged-branch auto-delete; a re-queued build reuses or resets the issue branch).

## Acceptance criteria

1. A build killed at any point after its first push leaves the branch on the remote at its last commit; the issue remains in the normal `status:building` → retry/groundskeeper reconciliation path (no PR exists).
2. Publish-step failure escalations name the branch and last pushed commit.
[...]

PR #692, opened 2026-07-25
Give the three PR-fixing loops the build worker's checkpoint step

swampratnz:
## Summary

Closes the *"the agent stalled and its committed work died with the runner"* hole in **autofix**, the **conflict resolver**, and the **revise** loop. The build worker got a deterministic checkpoint for exactly this failure (#633/#682); the three loops that push to existing PR branches never did.

**The prompt rule already exists, and is provably not enough.** Every one of these loops has an `EXECUTION MODEL` block telling the agent to run commands synchronously — the conflict resolver's says verbatim `NEVER end your turn to "wait" for a command.` Agents ignored it anyway:

| PR | Loop | The agent's own final words |
|----|------|------------------------------|
| #606 | revise | *"I'll wait for the monitor notification before continuing with the security test suite, build, and push."* |
| #609 | conflict | *"I'll stop issuing further commands now and wait for the running Monitor task to notify me when `npm run test:security` finishes."* |

Both had already **committed** work. Both lost it. Both PRs escalated `needs-human` having produced nothing — and #609's *"semantically incompatible"* conflict turned out to be a clean `main` merge a human finished in minutes. Adding more prompt text was not going to fix a rule the model already had, so this is the deterministic backstop instead.

## What changed

Each loop now runs the same post-agent checkpoint, placed **immediately before its existing verify step** so a recovery registers as the new commit it is:

- Runs only **after the agent has exited** — no process left to influence it or read the token — with the job's `GITHUB_TOKEN`, which outlives the agent's ~1h App token.
- **No-ops** when the agent committed nothing, or when it already pushed.
- Only ever **fast-forwards** the PR branch. If the remote moved underneath, the work is parked on a `-ckpt-<run_id>` ref rather than rewriting someone else's push.
- Leaves a marker-guarded comment stating plainly that the work **never cleared the agent's own gate**.

Each loop's checkpoint marker is distinct from its attempt/escalation markers, so a recovery notice can never inflate an attempt cap or read as an escalation.

## Security / privacy impact

No new tool, RBAC surface, or data path — CI/workflow plumbing only.

The token posture is unchanged and deliberate: the checkpoint step is the *only* place its `GITHUB_TOKEN` appears, it runs after the agent process is gone, and the checkouts keep `persist-credentials: false`, so there is no new window in which an agent reading untrusted PR/CI content could reach a credential. The push target is constructed from the workflow's own branch expression (passed via `env`, never `${{ }}`-interpolated into the script body, matching the existing convention in these files for refs that may contain shell metacharacters).

**Recovered work cannot be laundered into a merge.** It is unverified by construction, so the comment says so, CI runs on the push and adjudicates, and the auto-merge loop still requires all checks green *plus* a fresh `LGTM` from the review worker. The bounded attempt caps (2 for autofix/revise, 1 for conflict) are untouched.

⚠️ This PR touches `.github/`, `CLAUDE.md`, and `docs/PIPELINE.md` — all governance paths — so by design the auto-merge loop will not touch it and it needs a **human merge**.

## How verified

Workflow/docs only — no source or test changes, so the suite is unaffected.

- All three workflows **parse as valid YAML**, and the checkpoint step is confirmed to sit at the index immediately before each job's `Verify` step.
- The `run` body passes **`bash -n`** with the Actions expressions stubbed out.
- Both early-exit guards dry-run correctly: *agent committed nothing* and *agent already pushed* each no-op.
- `npm run lint`, `npm run format:check`, `npm run typecheck` — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QgYZcxN4CTBbCeLVTGu8aH)_

### DIFF
--- a/.github/workflows/pipeline-pr-autofix.yml
+++ b/.github/workflows/pipeline-pr-autofix.yml
@@ -322,7 +322,16 @@ jobs:
             echo "Agent committed nothing — nothing to checkpoint."; exit 0
           fi
           remote_sha="$(gh api "repos/${{ github.repository }}/git/ref/heads/${BRANCH}" --jq '.object.sha' 2>/dev/null || true)"
-          if [ "${head_sha}" = "${remote_sha}" ]; then
+          # A failed/404 gh api call can leave a NON-sha in remote_sha (an error
+          # body passed straight through --jq, e.g. the literal "null") — the
+          # same trap pipeline-build.yml had to harden after a real incident.
+          # Treat anything that is not a 40-hex sha as "remote tip unknown":
+          # blanking it skips the divergence pre-check and attempts the ordinary
+          # NON-FORCE push, which the server rejects unless it is a
+          # fast-forward. Without this a bogus value fails the ancestor test and
+          # diverts a perfectly fast-forwardable recovery onto a side ref.
+          if ! printf '%s' "${remote_sha}" | grep -qE '^[0-9a-f]{40}$'; then remote_sha=""; fi
+          if [ -n "${remote_sha}" ] && [ "${head_sha}" = "${remote_sha}" ]; then
             echo "Agent already pushed ${head_sha} — nothing to checkpoint."; exit 0
           fi
           target="${BRANCH}"
--- a/.github/workflows/pipeline-pr-autofix.yml
+++ b/.github/workflows/pipeline-pr-autofix.yml
@@ -84,6 +84,9 @@ jobs:
       # (a re-run of the same failed CI must not read one real attempt as two).
       ATTEMPT_MARKER: '<!-- pipeline-autofix-attempt -->'
       ESCALATE_MARKER: '<!-- pipeline-autofix-escalation -->'
+      # Its own marker so a recovered-work notice can never inflate the
+      # attempt count or be mistaken for an escalation.
+      CHECKPOINT_MARKER: '<!-- pipeline-autofix-checkpoint -->'
     steps:
       - name: Inert notice (token not set)
         if: env.HAS_TOKEN != 'true'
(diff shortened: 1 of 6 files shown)


swampratnz:
Both observations are correct; neither needs a code change, but you asked me to confirm the first is deliberate — it is.

**No resume-from-`-ckpt` mechanism: intentional, and I think resuming would be wrong here.** The build worker's #667 resume exists because a re-queued build re-runs against the same *issue* and genuinely wants to continue a half-built branch — there is no other writer.

These loops are the opposite case. The side ref is only ever created when the PR branch **moved underneath the agent** — someone (a human, or another loop) pushed while it worked. So by construction the checkpoint is built on a base that is no longer current, and auto-resuming from it would mean reapplying stale work over whatever landed in the meantime. The correct next action is to work from the *new* branch tip, which is exactly what the next attempt or the human does today. The side ref exists to make the lost work recoverable, not to be replayed automatically.

It's also the rare path: the common failure (the one that motivated this PR — #606/#609) is a stalled agent on a branch nobody else touched, which fast-forwards cleanly onto the PR branch and needs no resume at all.

**Token in `push_url`: deliberately left matching the build worker.** You're right that it's the accepted precedent rather than a new regression. I considered a credential helper or `http.extraheader`, but neither is clearly better here: the helper writes the token to disk, and `git -c http.extraheader=…` still puts it on the command line. Meanwhile the exposure window is a single `git push` on an ephemeral runner *after* the agent process has exited, so there is no consumer left to read `/proc`. Given that, diverging from `pipeline-build.yml`'s reviewed pattern would buy no real security while making four checkpoint implementations inconsistent — which is its own maintenance hazard. If the token handling is ever changed, it should change in all four together.

---

Issue #222, opened 2026-07-07
Bug (medium): PR-review verdict routing misroutes verdicts and posts spurious failures on cancelled runs


swampratnz:
From the 2026-07-07 deep review (confirmed). Both are in the review→revise routing added in #198; worth fixing together (same file).

**1. Verdict grep is too wide (`.github/workflows/pipeline-pr-review.yml:157`).** The post step does `head -n 6 review.txt | grep -qiE 'changes requested'` **before** `needs a human decision`. The mandated review format puts finding bullets from line 3, so reviewer narration is inside the window. If an LGTM verdict has a bullet like "Previous *changes requested* were all addressed," the first grep matches and dispatches revise; `pipeline-pr-revise.yml`'s precheck is strictly looser (tests the whole comment body) so it passes too, the agent finds nothing, pushes no commit, and the verify step labels the clean approved PR `needs-human` and fails red — the exact PR-#196-style stall the loop was built to prevent, now self-inflicted. Conversely a "Needs a human decision" verdict with an early "changes requested" bullet goes to revise instead of `needs-human`. **Fix:** anchor to the verdict line only (e.g. `sed -n '2p'`, or match a line starting with the phrase) and evaluate "needs a human decision" first; tighten the revise precheck to the same anchored form.

**2. Spurious failure comments on cancelled runs (`:115`).** The post step condition is `always() && HAS_TOKEN`, and `cancel-in-progress: true` cancels in-flight reviews on a new push. `always()` runs on cancellation → empty execution file → it posts "PR review (automated): no review output was captured… Failing the check" on a healthy PR every time two pushes land within a review's runtime. That comment also carries the literal `PR review (automated)` marker the revise precheck selects on. **Fix:** change the condition to `!cancelled() && HAS_TOKEN` (analyze already has `continue-on-error`), and give the deliberate no-output comment a first line that does not start with `PR review (automated)`.

One PR.

PR #749, opened 2026-07-27
Make the PR-review verdict a typed artifact, not prose three loops re-parse

swampratnz:
## Summary

Applies the "artifact contract at every handoff" principle to the pipeline's most-parsed handoff, and fixes a **live latent bug** found while doing it.

### The bug

Three workflows consume a PR-review verdict, each with its own regex over the same free prose:

| Consumer | Filter | Has #731's fix? |
|---|---|---|
| `pipeline-pr-review.yml` (routes) | `grep -vE '^[[:space:]]*[-*][[:space:]]'` | yes |
| `pipeline-pr-revise.yml` (re-verifies) | jq equivalent | yes |
| `pipeline-pr-automerge.yml` (gates merges) | `grep -vE '^[[:space:]]*[-*]'` | **no** |

#731 fixed "a bolded `**Changes requested**` is not a markdown bullet" in two of the three. Auto-merge kept the old filter, so a review whose verdict the model renders in bold — `**LGTM, looks good and ready for a human to merge**` — matches `^\s*[-*]`, gets dropped, leaves `verdict` empty, and the gate concludes "not an LGTM approval" and skips. **A fully-approved PR would sit unmerged forever, with no error on the PR, in the logs, or anywhere else.** Verified against the real filter before changing anything:

```
extracted verdict: [PR review (automated):]
RESULT: would SKIP — bolded LGTM is invisible to auto-merge
```

### The fix

A contract, not a better regex:

- The review model is asked to emit `<!-- verdict:LGTM -->` / `<!-- verdict:CHANGES_REQUESTED -->` / `<!-- verdict:NEEDS_HUMAN -->` on line 2. It renders invisibly, so the comment still reads naturally to humans.
- `pipeline-pr-review.yml` — the only place a review comment is composed — decides the verdict **once** (token if present, else the prose fallback), strips any whole-line token the model wrote, and stamps exactly one **authoritative** token after the `PR review (automated):` marker. The stamp is written by the workflow, never by the model.
- Consumers read that token. The authoritative one is always first, so a review that legitimately quotes a token mid-sentence — reviews of this very machinery do — can neither be mangled nor mistaken for the verdict.
- The prose fallback stays for comments posted before the contract existed, but is now **shared** rather than reimplemented three times, which is what fixes auto-merge's missing #731 patch.
- Routing no longer re-parses the posted text, so the #222 decoy class (a finding bullet reading like a verdict) is structurally impossible on the token path.
- An unrecognisable verdict routes nowhere — no label, no dispatch, nothing for auto-merge to approve — preserving today's fail-safe behaviour: it stalls visibly rather than guessing.

### Why not a shared script

`pipeline-pr-review.yml` runs on `pull_request` and therefore checks out **PR-controlled code**, in a job holding `pull-requests: write` and `actions: write`. Calling a repo script from that checkout would let a malicious PR rewrite the parser. The logic stays inline shell in all three workflows, and drift is prevented by a test instead.

## Tests

`tests/reviewVerdict.test.ts` (new, 12 tests) pins both halves:

- **Drift**: both shell helpers must be byte-identical across all three consumers — the exact failure that caused this bug — compared after dedenting so YAML nesting doesn't matter.
- **Behaviour**: the helpers are executed via `bash` against fixtures covering a plain token, all three verdicts, lowercase/extra-whitespace token drift, a missing token, a decoy token quoted in a finding, the **bolded-LGTM** case that stalled auto-merge, and the #222 finding-bullet decoy.

## Verification

- `npm run typecheck`, `npm run lint`, `npm run format:check` — clean
- `npm test` — 2680 tests, **0 failures**
- `npm run test:security` — 989/989, manifest updated in this diff
- `npm run build` — clean
- The review workflow's stamping logic simulated end-to-end in bash against the three real-world cases (model emits token / model omits token with bolded LGTM / decoy token), confirming one authoritative token lands on line 2 in each

## Note for the reviewer

This PR modifies `pipeline-pr-review.yml`, so its own automated review will likely post "no output captured this run" — the documented self-modification case (the action runs the default branch's workflow version). That is expected here and not a failure; it reviews normally once merged.

Touches `.github/workflows/`, so a human merge is required by design.

https://claude.ai/code/session_017W3YHSwScRhzxhp8YuP5Q7

---
_Generated by [Claude Code](https://claude.ai/code/session_017W3YHSwScRhzxhp8YuP5Q7)_

### DIFF
--- a/.github/workflows/pipeline-pr-automerge.yml
+++ b/.github/workflows/pipeline-pr-automerge.yml
@@ -269,20 +269,41 @@ jobs:
             review_body="$(jq -r '.body' <<<"$latest_review")"
             review_at="$(jq -r '.createdAt' <<<"$latest_review")"
 
-            # Verdict = first 4 lines minus bullet lines, so a phrase inside a
-            # finding can't be misread as the verdict. The comment is GUARANTEED
-            # marker-first: the review workflow deterministically trims any
-            # agent preamble at post time (see pipeline-pr-review.yml), so the
-            # strict ^-anchored match above and this top-of-comment parse are
-            # unambiguous even when review commentary QUOTES the marker string
-            # further down (PR #656 review found substring-based extraction
-            # here was decoy-prone in both directions; anchoring the SHAPE at
-            # the source removes the ambiguity class entirely).
-            verdict="$(printf '%s' "$review_body" | head -n 4 | grep -vE '^[[:space:]]*[-*]' || true)"
-            if printf '%s' "$verdict" | grep -qiE 'changes requested|needs a human decision' \
-               || ! printf '%s' "$verdict" | grep -qiE 'lgtm|ready for a human to merge'; then
-              echo "PR #$n: latest review verdict is not an LGTM approval — skip."
-              echo "::endgroup::"; continue
+            # Verdict via the canonical token the review workflow stamps on
+            # line 2 (see pipeline-pr-review.yml's "canonical verdict contract").
+            # This replaced a local prose regex that had DRIFTED from the other
+            # two consumers: it lacked #731's fix, so a review whose verdict the
+            # model rendered in bold (`**LGTM, ...**`) matched the bullet filter,
+            # left `verdict` empty, and this gate skipped a fully-approved PR
+            # forever. Reading one exact token removes that class — the token is
+            # written deterministically by the workflow, never by the model, and
+            # a review that QUOTES a token further down can't win because the
+            # authoritative one is always first.
+            # Keep both helpers identical across the three verdict consumers;
+            # tests/reviewVerdict.test.ts fails if any copy drifts.
+            canonical_verdict() {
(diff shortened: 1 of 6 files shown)


swampratnz:
The red `review` check here is expected and not a defect — recording it on the thread since it has now fired twice and is the only non-green check on this PR.

This PR modifies `pipeline-pr-review.yml`, and for a `pull_request` event the review action runs the workflow version from the **default branch**, so it cannot review a change to itself. The workflow deliberately fails the check in that situation rather than staying silent, which is correct behaviour — the alternative is a review that quietly produces nothing and looks green. PR #731 hit exactly the same thing for the same reason and was merged as-is; this one will review normally once merged.

Everything else on this PR is green: `build`, `lint`, `security-invariants`, `discover`.

Local verification on this head: typecheck, lint, format, build clean; `npm test` 2680 tests with 0 failures; `npm run test:security` 989/989 with the manifest updated in the diff; and the review workflow's stamping logic simulated end-to-end in bash against the three real cases (model emits the token, model omits it with a bolded LGTM, decoy token quoted lower down), confirming exactly one authoritative token lands on line 2 in each.

---

Issue #719, opened 2026-07-26
Deflake the shared-table aggregate tests: usageStats asserts exact values over whole-table scans that parallel test files contaminate — same class #704 fixed for the context builder


swampratnz:
_VISION theme: **Reliability & ops** (`theme:reliability`). Originally filed as an evidence-only `community-feedback` note during the 2026-07-26 pipeline watch; relabelled into the proposal lane at the owner's direction — the pipeline has no intake for evidence-only issues._

## Problem statement

**Who it helps:** every PR author (bot or human) and the shared Max pool. A flake in this class lands on **unrelated PRs**, costs a ci-retry rerun plus potentially an autofix agent engagement, and corrodes trust in red CI.

**The class:** a test seeds rows, calls a function that aggregates over the **whole table**, and asserts exact values. Test files run in parallel against one Postgres service container, so any concurrently-running file inserting rows into the same table changes the aggregate.

**Evidence — two live instances on 2026-07-26, both verified from CI logs:**

1. `tests/contextBuilder.test.ts` failed on the docs-only PR #702: `runContextBuilder` clustered over the whole `interactions` table and `tests/knowledgeCandidateDedupEmbedCount.test.ts`'s rows (same one-hot axis 31, cosine 1.0) could win the `maxSummaries=1` slot. **Fixed in #704** via an injectable corpus scoped to the file's own `RUN` prefix. Reproduced deterministically before fixing.
2. `tests/repository.test.ts` `usageStats` tests failed on the unrelated te reo PR #717 (run 30204969555): `costByRole` counted 5 rows where the test seeded 1, and `byPlatform` returned `1.2000000000000002 !== 1.2` **after** its own `retryOnSharedTableInterference` helper ran — the helper detects churn but cannot fix a *stable* contaminated sum. **Still exposed**; the autofix loop burned an engagement treating it as a flake.

## Proposed approach

Apply #704's precedent to the `usageStats` tests, plus a bounded audit:

1. **`usageStats` tests** (`tests/repository.test.ts`): scope the assertions to the test's own rows. `usageStats` is legitimately table-wide in production, so mirror #704's seam: either an injectable row-predicate for tests, or (lighter) assert **deltas** — capture the aggregate before seeding, seed, assert the difference — which is monotone-safe under concurrent inserts only if foreign inserts can't land between the two reads; where that race is real, prefer the injectable predicate.
[...]

PR #750, opened 2026-07-27
Measure the pipeline loops: reconstruct a per-loop outcome ledger

swampratnz:
## Summary

The pipeline has seven agent loops that fix, revise, resolve and merge PRs, and **nothing measured any of them**. A loop could burn an escalation on a one-command fix, die mid-turn without pushing, or thrash for weeks, and the only trace was a marker comment on a single PR that nobody aggregates. "Is this loop earning its tokens?" had no answer, so every change to a loop was argued from anecdote.

This adds the missing evidence base without adding any new state.

### How it works

`scripts/pipeline-outcomes.mjs` reconstructs each loop's record from evidence that **already exists** — the marker comments every loop posts when it engages, checkpoint-recovers, or escalates. No new writes to any PR, no new state to keep in sync, no LLM, no Max pool. Same read-only deterministic shape as `scripts/check-changelog-coverage.mjs`.

`pipeline-outcomes.yml` runs it weekly, always publishes to the run summary (free, no notification noise), and keeps **one** tracking issue in sync only when a loop failed to finish on its own — auto-closing on a clean window, the same self-clearing contract as `changelog-coverage.yml`, so a quiet pipeline stays quiet.

### Reading the report

Run against real markers from the last day, it produces:

| Loop | Engaged | Recovered (agent stopped early) | Escalated to human | Clean |
| --- | --- | --- | --- | --- |
| autofix | 1 | 0 (0%) | 1 (100%) | 0 (0%) |
| revise | 2 | 0 (0%) | 2 (100%) | 0 (0%) |
| conflict-resolver | 2 | 1 (50%) | 0 (0%) | 1 (50%) |

**Recovered** is the column that matters: an agent that committed work and then ended its turn without pushing, rescued by that loop's deterministic checkpoint step. Each one is a prompt/harness defect in the loop — not a code defect in the PR — and it is the failure mode that has cost the most here (#606 and #609 both escalated `needs-human` with finished work stranded on the runner). **Escalated** is the loop correctly giving up.

Recovered and Escalated are deliberately **not** mutually exclusive — one engagement can be recovered *and then* escalated — so those percentages are shares of engagements rather than slices of a pie and may sum past 100%. Splitting them into exclusive buckets would hide the double-failure case, which is the one most worth seeing.

## Security posture

Reads PR titles/bodies/comments as jq/node data only and runs no PR-controlled code — the script comes from the checked-out default branch, never a PR head (the workflow triggers only on `schedule`/`workflow_dispatch`). Permissions are `contents: read`, `issues: write`, `pull-requests: read`.

**Marker authenticity** is the one real trust boundary, since markers are plain text in a public thread and decide both the numbers and whether the tracking issue auto-closes. Markers count only from `github-actions`/`github-actions[bot]` (both `gh` renderings — GraphQL and REST differ, and matching one makes the gate silently match nothing). `claude[bot]` is excluded deliberately: every real marker is written by a deterministic `GITHUB_TOKEN` step, while the revise agent uniquely holds `Bash(gh pr comment:*)`, runs under that identity, and reads prompt-injectable PR content — so admitting it would let an injected agent fabricate rows, which is worse than no gate because the gate implies the rows are trustworthy. Recorded in `docs/SECURITY.md`.

The `window_days` dispatch input reaches the steps via `env:` (never spliced into a `run:` body) and is validated to a positive integer in 1..365 before use.

## Tests

`tests/pipelineOutcomes.test.ts` — **15 tests**, two of them `SECURITY:`-prefixed (`security-floor.json` updated to match):

- Counting rules: per-loop engaged/recovered/escalated, the notable-PR list and ordering, window filtering and `--window-days`, an escalation-only PR still counting as engaged, auto-merge's blocked notice as friction.
- Regressions from review: `Clean` counted per-PR so a sibling double-failure can't cancel a genuinely clean engagement; a single engagement that both recovered and escalated counting as one failed engagement; non-numeric window falling back to the default instead of "NaN days".
- **SECURITY**: a marker from an arbitrary commenter contributes nothing; a marker attributed to `claude`/`claude[bot]` contributes nothing. Plus a test that every accepted `github-actions` rendering *is* recognised, so the gate can never silently zero out a row.
- Robustness: malformed stdin exits 0 rather than failing the caller; a PR with no `comments` field.

## Verification

- `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run build` — clean
- `npm test` — full suite green apart from the known shared-table flake in `repeatQuestionShortcutRouter` (issue #719 class), which passes in isolation and which this diff cannot affect: there is no `src/` change here at all
- `npm run test:security` — manifest matches, updated in this diff for the two new `SECURITY:` tests

Touches `.github/workflows/`, `scripts/` and `docs/SECURITY.md`, so a human merge is required by design.

https://claude.ai/code/session_017W3YHSwScRhzxhp8YuP5Q7

### DIFF
--- a/.github/workflows/pipeline-outcomes.yml
+++ b/.github/workflows/pipeline-outcomes.yml
@@ -1,0 +1,130 @@
+name: Pipeline outcomes
+
+# Weekly self-measurement for the agent loops (see scripts/pipeline-outcomes.mjs).
+#
+# The pipeline has seven loops that fix, revise, resolve and merge PRs, and
+# nothing measured any of them: a loop could escalate on a one-command fix, die
+# mid-turn without pushing, or thrash for weeks, and the only trace was a marker
+# comment on one PR that nobody aggregates. "Is this loop earning its tokens?"
+# had no answer, so every change to a loop was argued from anecdote.
+#
+# This reconstructs the record from evidence that ALREADY exists — the marker
+# comments the loops post when they engage, checkpoint-recover, or escalate. No
+# new state, no writes to any PR, no LLM, no Max pool: the same read-only
+# deterministic shape as changelog-coverage.yml, and like it, the tracking issue
+# auto-closes when there is nothing to act on.
+#
+# Reads PR titles/bodies/comments as jq data only and runs no PR-controlled
+# code — `scripts/pipeline-outcomes.mjs` comes from the checked-out default
+# branch, never from a PR head.
+
+on:
+  schedule:
+    - cron: '23 20 * * 1' # 20:23 UTC Monday (~08:23 NZ Tuesday)
+  workflow_dispatch:
+    inputs:
+      window_days:
+        description: 'Days of history to report on (default 14).'
+        type: string
+        required: false
+
+permissions:
+  contents: read # checkout + run the report script
+  issues: write # open/update/close the single tracking issue
+  pull-requests: read # gh pr list / gh pr view
+
+concurrency:
+  group: pipeline-outcomes
+  cancel-in-progress: true
+
+env:
+  ISSUE_TITLE: '📊 Pipeline loop outcomes: loops that did not finish on their own'
+  WINDOW_DAYS: ${{ inputs.window_days || '14' }}
+  # Bounds the API calls: PRs updated longer ago than the window can hold no
+  # in-window markers, so there is nothing to gain by fetching them.
+  SCAN_LIMIT: '80'
+
+jobs:
+  report:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false # no git credential needed after checkout
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 24
+
+      - name: Collect PR comment history
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          cutoff="$(date -u -d "-${WINDOW_DAYS} days" +%Y-%m-%dT%H:%M:%SZ)"
+          gh pr list --repo "$GITHUB_REPOSITORY" --state all --limit "$SCAN_LIMIT" \
+            --json number,updatedAt \
+            --jq "[.[] | select(.updatedAt >= \"$cutoff\") | .number] | .[]" > numbers.txt
+          echo "Scanning $(wc -l < numbers.txt) PR(s) updated since $cutoff."
+          : > prs.ndjson
+          while read -r n; do
+            [ -n "$n" ] || continue
+            gh pr view "$n" --repo "$GITHUB_REPOSITORY" \
+              --json number,title,url,createdAt,comments >> prs.ndjson
+          done < numbers.txt
+          jq -s '.' prs.ndjson > prs.json
+
+      - name: Build the report
+        id: report
+        run: |
+          set -euo pipefail
+          node scripts/pipeline-outcomes.mjs --window-days "$WINDOW_DAYS" < prs.json > report.md
+          cat report.md
+          # Always publish to the run summary — free, and no notification noise.
+          cat report.md >> "$GITHUB_STEP_SUMMARY"
+          # Only nag when a loop actually failed to finish on its own; a clean
+          # window closes the issue instead (same self-clearing contract as
+          # changelog-coverage.yml).
+          if grep -q '^### PRs where a loop did not finish' report.md; then
+            echo "actionable=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "actionable=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Open or update the tracking issue
+        if: steps.report.outputs.actionable == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          {
+            cat report.md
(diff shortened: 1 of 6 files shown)


swampratnz:
Addressed in `c7206c9` — but the blocking finding is a **false positive**, and the distinction matters enough to record rather than quietly "fix".

**The command injection does not exist.** The claim is that `cutoff="$(date -u -d "-${WINDOW_DAYS} days" ...)"` executes a `$(...)` payload in `WINDOW_DAYS` because "Bash performs command substitution inside double-quoted strings". Bash performs command substitution on the **source text** of the script, not on the **result** of a parameter expansion — expansion results are never re-scanned for further substitution. `WINDOW_DAYS` reaches the step as an environment variable (the `${{ }}` is in an `env:` block, never inside a `run:` body), so its value is inert data. Verified empirically with the exact line:

```
$ export WINDOW_DAYS='$(touch /tmp/PWNED)$(echo INJECTED)'
$ cutoff="$(date -u -d "-${WINDOW_DAYS} days" +%Y-%m-%dT%H:%M:%SZ)"
date: invalid date '-$(touch /tmp/PWNED)$(echo INJECTED) days'
$ [ -f /tmp/PWNED ] && echo executed || echo "no execution — literal data"
no execution — literal data
```

GitHub's documented script-injection hazard is `${{ … }}` interpolated **directly into a `run:` body**, which splices attacker-controlled text into the script source before bash parses it. This workflow does not do that anywhere — the env-var indirection used here *is* the recommended mitigation. That also means the sibling workflows are not carrying a latent instance of this, and shouldn't be changed on this basis.

**What I did change**, because the *consequences* the review describes are real even though the mechanism isn't:

- A `Validate the window` step rejects anything that isn't a positive integer in 1..365, failing fast with a clear message instead of an opaque mid-job `date` error. Kept as defence-in-depth so a future refactor toward a `run:`-body interpolation is already covered, with a comment recording why it is *not* an injection fix.
- The script falls back to the default window instead of rendering `last NaN days` and matching nothing (your minor point 1) — new test pins it.
- Both the code and the report footer now state that **Recovered** and **Escalated** are shares of engagements, not slices of a pie (your minor point 2). They are deliberately non-exclusive: one engagement can be checkpoint-recovered *and then* escalated — #609 was exactly that — and splitting them into exclusive buckets would hide the double-failure case, which is the one most worth seeing.

Verification on the pushed commit: typecheck, lint, format, build clean; `tests/pipelineOutcomes.test.ts` 10/10; validation step exercised against `14` (accept), `abc`, `$(touch …)`, `0`, `400` (all rejected).

Thanks for the marker-constant cross-check against the four real loop workflows — that was the failure mode I most wanted a second pair of eyes on, since a stale marker would silently zero out a row.

---

swampratnz:
Both "minor, non-blocking" notes are fixed in `36fc16f` rather than deferred — they're the same class as the `Clean` bug this branch already fixed: a *measurement* tool silently reporting wrong numbers, which is the one defect category this PR exists to eliminate.

**`gh pr list` ordering** — the substantive one. It sorts by **creation** date, so the newest-80 slice would drop an older PR that a loop touched recently, which is exactly the PR whose markers matter, and the `updatedAt` filter applied afterwards couldn't recover it. At this repo's throughput (750+ PRs; ~20 merged in the last day) that's a live undercount, not a theoretical one. Now `--search "sort:updated-desc"`, so the slice matches the filter that follows it.

**Marker authenticity** — markers are plain text in a public thread, and since this report now drives an auto-closing tracking issue, a spoofed row isn't merely cosmetic. Markers are now counted only from the pipeline's own bot identities. Both `gh` renderings are accepted for exactly the reason `pipeline-pr-automerge.yml` accepts both — GraphQL renders `github-actions`, REST renders `github-actions[bot]`, and matching only one previously made an identity gate match nothing. Given a stale-identity list would silently zero out a row (the failure mode your marker cross-check on the first pass was guarding against), that risk is now pinned by a test asserting every accepted rendering is recognised, alongside a `SECURITY:` test that a marker from an arbitrary commenter contributes nothing.

Tests now 14; `security-floor.json` updated in the same diff for the one new `SECURITY:` test. Typecheck, lint, format, and build clean.

Thanks for both catches — the `Clean` finding in the previous round was a genuine bug that would have inverted the tool's headline signal, and the ordering issue would have quietly under-reported failures.

---

swampratnz:
Correct, and this was my error — fixed in `2bf2162`. I widened the allowlist "to be safe against a loop that posts via the App token" without checking whether any loop actually does. Verified both halves of your claim before changing anything:

**No loop posts a marker as `claude`.** Every marker — autofix `ATTEMPT`/`ESCALATE`, revise `ATTEMPT`/`ESCALATE`/`CHECKPOINT`, conflict `CONFLICT`/`CHECKPOINT`, auto-merge `READY`/`BLOCKED` — is written by a deterministic step under `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}`, so it always lands as `github-actions`.

**The revise agent is uniquely exposed.** It holds `Bash(gh pr comment:*)` — `pipeline-pr-autofix.yml` and `pipeline-pr-conflict.yml` grant no comment verb at all — runs under `claude[bot]`, and reads untrusted PR content. So accepting `claude[bot]` handed the single injectable identity in the pipeline the ability to fabricate ledger rows.

That makes it worse than having no identity gate, which is the part worth stating plainly: an unauthenticated ledger is merely unreliable, but a gate that admits the injectable identity is unreliable *while implying the rows can be trusted*. The previous commit's claim that spoofing was closed would have been false.

The allowlist is now the two `github-actions` renderings only. Added a `SECURITY:` test that a `claude[bot]`-attributed marker contributes nothing, and narrowed the rendering test to the identity that actually posts. Tests 15; `security-floor.json` updated in the same diff (now 2). Typecheck, lint, format, build clean.

Three genuine findings across three passes on this PR — the `Clean` undercount, the creation-date scan ordering, and now this. Appreciated.

---

PR #796, opened 2026-07-28
Build worker: deterministic lane ownership, drop the agent's gh issue edit grant (audit N4)

swampratnz:
## Summary

Closes the **N4** finding from the 2026-07-28 audit — the last remaining actionable item, which I deferred from PR #794 because it needed lane-ownership restructuring rather than a one-line removal.

The build agent held `Bash(gh issue edit:*)` only to move its *own* issue through the status lanes, but that matcher cannot pin the issue number — so an injected build agent could run `gh issue edit <any> --add-label status:approved` and spawn a full build of an arbitrary, never-adversarially-reviewed issue (an App-token label add *does* re-trigger the build workflow), or strip `needs-human` off an escalated one. The grant is now **removed entirely** and the workflow owns every lane transition deterministically with its `GITHUB_TOKEN` (which cannot re-trigger a workflow):

- **Claim step** (new, before the agent runs) marks `status:building` and removes `status:approved` — idempotent across retries, guarded on `needs-human`, and posts the claim comment only on the first attempt.
- **Verify step** marks `status:built` (removing `status:building`) once it has *confirmed* the PR, guarded on `needs-human` to preserve the one-lane invariant.
- **Deliberate refusal** (infeasible/unsafe proposal) is signalled by the agent writing a git-ignored `needs-human.md` file — the same file-signal pattern as the existing handoff note, read from the workspace rather than from untrusted comment text. The verify step turns it into the `needs-human` label, clears the lane, posts the agent's own explanation, and exits 0 so the retry loop does not re-run a proposal the agent already judged unbuildable.

The agent prompt is updated (lane labels are not the agent's to set; refusal writes the file, a merely-unmakeable-green gate is a comment only). The pre-existing recovery and final-attempt escalation paths already used deterministic `gh issue edit` and are unchanged. `docs/PIPELINE.md` records the rationale.

## Scope (important — this closes N4 in the Action lane only)

N4's exploit shape lives in the **unattended, injection-exposed Action lane** (`pipeline-build.yml`), and that is what this PR closes. The audit scopes N4 to that workflow's `allowedTools`. The **optional fallback Build *Routine*** (a live `/loop` session, listed in the mapping table as "Routine hourly as fallback") is a *different execution model*: it is not bounded by `pipeline-build.yml`'s `--allowedTools`, and it must self-manage its `status:*` labels because it dynamically picks which approved issue to claim — so it inherently cannot be given a per-run-pinned grant. If that fallback is run it retains the un-pinned relabel capability; it is the higher-trust, human-operated lane and should be run attended. This PR documents that scope (in the least-privilege bullet and next to the mapping table) rather than claiming N4 is eliminated across both execution models. (Thanks to the automated review for catching the overclaim in the first version of this description.)

## Security / privacy impact

Net hardening of the **Action lane**. Removes the agent capability that let the unattended build worker label an *arbitrary* issue and thereby inject unreviewed work into the build lane. All label writes in that lane now originate from the workflow's `GITHUB_TOKEN`, which structurally cannot re-trigger the build workflow, so the "approve → build" chain can no longer be driven by a compromised agent turn. The refusal signal is a file the agent writes (not comment text), keeping it off the untrusted-comment channel; an injected agent that writes it only escalates to a human (fail-safe). The fallback Build Routine is out of scope (see above). No change to the merge gate, RBAC, or any runtime bot behaviour — this is pipeline-internal.

## How verified

Workflow/doc-only change (no `src/`), so the TypeScript gates are unaffected from the current `main` baseline. Verified:

- `python3 -c yaml.safe_load` parses `pipeline-build.yml`; `npm run format:check`, `npm run lint`, `npm run context:check` all exit 0.
- Manually traced every path: claim (attempts 1–3, idempotent, needs-human-guarded), success built-transition, deliberate refusal (exit 0, no retry, recovery skipped), gate-failure (no `needs-human.md` → existing retry/escalation intact), and confirmed the removed grant is not referenced by any test or the context pack.
- The workflow triggers on `issues.labeled`, not `pull_request`, so it does not self-execute on this PR.
- The first `build` check failed on a transient `onnxruntime-node` binary-download timeout during `npm ci` (unrelated to the diff); `ci-retry.yml` re-runs that automatically.

Because behavioural verification of a pipeline workflow only happens on a live build, the first real build after merge is worth a glance. Governance path (`.github/`), so it needs a human merge.

### DIFF
--- a/.github/workflows/pipeline-build.yml
+++ b/.github/workflows/pipeline-build.yml
@@ -234,6 +234,36 @@ jobs:
           fi
           echo "hint=${hint}" >> "$GITHUB_OUTPUT"
 
+      # Claim the issue into the building lane DETERMINISTICALLY, before the
+      # agent runs — the build agent no longer holds `gh issue edit` (audit
+      # 2026-07-28 N4), so the workflow owns every lane transition now. Uses the
+      # job's GITHUB_TOKEN, which (unlike the agent's claude[bot] App token)
+      # cannot re-trigger any workflow, so this label change starts nothing.
+      # Idempotent across retries: `--add-label` on an already-building issue is
+      # a no-op, and the claim comment is posted only on the first attempt so a
+      # rerun doesn't spam. Guarded on needs-human: a human (or the refusal path
+      # below) may have escalated the issue between attempts, and re-claiming it
+      # would silently pull it back out of the human lane — so if needs-human is
+      # set, do not touch the labels.
+      - name: Claim the issue (status:approved -> status:building)
+        if: env.HAS_TOKEN == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          labels="$(gh issue view "${ISSUE_NUMBER}" --json labels --jq '[.labels[].name] | join(",")' 2>/dev/null || true)"
+          case ",${labels}," in
+            *,needs-human,*)
+              echo "Issue #${ISSUE_NUMBER} is needs-human — not claiming (a human owns it)."
+              exit 0 ;;
+          esac
+          gh issue edit "${ISSUE_NUMBER}" --add-label status:building --remove-label status:approved || true
+          if [ "${{ github.run_attempt }}" = "1" ]; then
+            run_url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+            printf '🤖 Build worker claimed this issue (\`status:building\`). Implementing on a branch; a pull request will follow.\n\nRun: %s\n' \
+              "${run_url}" | gh issue comment "${ISSUE_NUMBER}" --body-file - || true
+          fi
+
       - name: Build the approved proposal
         id: build
         if: env.HAS_TOKEN == 'true'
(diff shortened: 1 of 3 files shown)


swampratnz:
Good catch on the overclaim — agreed, and addressed in `9471e77`.

The fix genuinely closes N4 in the **Action lane** (`pipeline-build.yml`), which is where the finding's exploit shape lives: an *unattended*, injection-exposed agent whose App-token label edit re-triggers a build. The **fallback Build Routine** is a separate execution model — a live `/loop` session, not bounded by that workflow's `--allowedTools`, and it *must* self-manage its `status:*` labels because it dynamically picks which approved issue to claim, so it can't be given a per-run-pinned grant. I didn't try to make it deterministic (it can't be); instead I documented the scope so nobody merges this believing both lanes are covered:

- The least-privilege bullet and the "Recommended mapping" note in `docs/PIPELINE.md` now state plainly that the deterministic lane ownership is the Action lane, and that the fallback Routine retains the un-pinned relabel capability and is the higher-trust, **run-it-attended** lane.
- The PR's Summary and Security/privacy sections are re-scoped accordingly, with an explicit "Action lane only" section.

Everything else in your review matched my own trace of the claim/verify/refusal/recovery paths. The one red check (`build`) was a transient `onnxruntime-node` binary-download timeout in `npm ci`, unrelated to the diff — `ci-retry.yml` re-runs it.

---

PR #812, opened 2026-07-28
Split repository.ts: extract admin audit + shortcut hits (audit L14, 5/N)

swampratnz:
## Summary

Fifth slice of the `repository.ts` carve-up (audit **L14**), continuing #804, #807, #809, #810.

Unlike the previous four, these two domains are **consumed by code that stays in `repository.ts`**, so they exercise the explicit-import half of the facade pattern:

- **`repository/adminAudit.ts`** — `recordAdminAction`, the `admin_audit` trail. `autoEnrollMemberWithAudit` (still in `repository.ts`) calls it *inside its transaction*, so `repository.ts` imports it explicitly.
- **`repository/shortcutHits.ts`** — the zero-model-call shortcut counters. `usageStats` (still in `repository.ts`) folds `sumShortcutHits` into its digest, so that's imported explicitly too.

`export *` re-exports for *callers* but does not bind names in the re-exporting module's own scope — the same subtlety #804 hit with `getResponseStyle` and #809 hit with `KNOWLEDGE_SEARCH_RELEVANCE_THRESHOLD`.

`repository.ts` is now **6,135 lines** — down from **7,137** at campaign start (−1,002, **14%**).

## One deliberate non-motion change

The private `Queryable` type alias (`Pick<PoolClient, 'query'>`) moves to `repository/shared.ts` and is now **exported**. `recordAdminAction` takes it, and two functions still in `repository.ts` do too, so it has to be shared. The alternative was duplicating the declaration — which is the debt this campaign exists to reduce.

The alias itself is **unchanged**; only its location and visibility. It's types-only, so there's no runtime surface change — but I'm calling it out explicitly rather than letting a "pure move" claim quietly cover it.

## Security / privacy impact

None. No SQL, logic, or signature change; no new dependency; no altered tool gating, identity derivation, or outbound filtering. `recordAdminAction` keeps its optional-`Queryable` shape, so the auto-enrol audit row still commits atomically with the member grant it describes. Neither domain carries a conversation-scoped admin read, so the 🔒 admin-scoping invariant is untouched. `tests/security-floor.json` unchanged: no `SECURITY:` test added or removed.

## How verified

All gates green locally against current `main`:

- `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run context:check` — all exit 0.
- `npm test` — **2114 pass / 0 fail** / 721 skipped (DB-gated; CI is the real verdict).
- `npm run test:security` — exit 0, **1047** SECURITY tests matching the manifest.
- **Byte-identical check:** all 3 moved functions compared whitespace-normalised against `origin/main` — **3 IDENTICAL, 0 differing**; `Queryable`'s alias text confirmed preserved.

Touches the 🔒 storage layer, so it needs a human merge.

---

**Campaign progress:** #804 → #807 → #809 → #810 → this PR. `repository.ts` 7,137 → 6,135.

**Remaining, in increasing order of care:**
1. **Member projects** — screens clean, but `tests/tools.test.ts` asserts the `// --- Member projects` banner exists *in `repository.ts`*; its own PR with that test edit as the focus.
2. **Background job costs** — coupled both ways (it calls `usageStats`), so it wants the same explicit-import treatment as this PR.
3. **The large contiguous blocks** — interactions/memory, knowledge (~1,200 lines, likely worth sub-splitting), members/identity-linking, moderation.
4. **The ~16 conversation-scoped 🔒 admin-read functions** — last, in their own dedicated PRs with the SQL scoping called out for explicit review.
5. **`tools.ts`** — needs its 4,577-line shared closure (`requireConfirm`, `callerScope`, `audited`, the rate-limit maps) extracted first; deliberately not started pending discussion, since it's the tool-gating spine.

**Still outstanding:** the cross-file test-pollution fix offered in #810's body. It reddened #809 (passed unchanged on retry) and forced a retry on `main`. It matters more from here, because the knowledge domain touches exactly the tables those global-matching assertions read.

---
_Generated by [Claude Code](https://claude.ai/code/session_019B8WC3ksmJrDbczjRK9584)_

### DIFF
--- a/src/storage/repository.ts
+++ b/src/storage/repository.ts
@@ -335,8 +335,6 @@ export async function recentConversationTail(
   }
 }
 
-/** A pooled connection or a transaction client — both expose `query`. */
-
 /**
  * Invalidate every context digest whose provenance refs include any of
  * `interactionIds`, deleting its still-*pending* knowledge candidates first
--- a/src/storage/repository.ts
+++ b/src/storage/repository.ts
@@ -35,13 +35,17 @@ export * from './repository/shared.js';
 export * from './repository/devTeamWatches.js';
 // `export *` does not bind names into this module's own scope, and the
 // knowledge queries below use this floor directly — so import it explicitly.
-import { KNOWLEDGE_SEARCH_RELEVANCE_THRESHOLD } from './repository/shared.js';
+import { KNOWLEDGE_SEARCH_RELEVANCE_THRESHOLD, type Queryable } from './repository/shared.js';
+import { recordAdminAction } from './repository/adminAudit.js';
+import { sumShortcutHits } from './repository/shortcutHits.js';
 export * from './repository/accessRequests.js';
 export * from './repository/contextDigests.js';
 export * from './repository/memberDiscovery.js';
 export * from './repository/docsIngestFailures.js';
 export * from './repository/policies.js';
 export * from './repository/roster.js';
+export * from './repository/adminAudit.js';
+export * from './repository/shortcutHits.js';
 
 // `export *` re-exports for CALLERS but does not bind the names in this
 // module's own scope, so anything still living here that uses an extracted
@@ -332,7 +336,6 @@ export async function recentConversationTail(
 }
 
 /** A pooled connection or a transaction client — both expose `query`. */
-type Queryable = Pick<PoolClient, 'query'>;
 
 /**
  * Invalidate every context digest whose provenance refs include any of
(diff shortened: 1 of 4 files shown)


swampratnz:
Fair catch — fixed in `9718c9e`.

Moving `type Queryable` into `shared.ts` took the declaration but left its one-line doc comment behind in `repository.ts`, documenting nothing. You're right that it's pure debris rather than a purposeful duplicate, and right that it contradicts this campaign's premise: an extraction that leaves residue isn't a clean extraction.

I also swept `repository.ts` for any other comment block orphaned by the four earlier extractions (a docblock followed by a blank line and then another docblock or a section banner) — **none found**, so this was the only one.

Gates re-run after the fix: `typecheck` / `lint` / `format:check` / `context:check` all exit 0, `npm test` 2114 pass / 0 fail, `test:security` 1047 matching the manifest.

---

Issue #924, opened 2026-07-31
Admins have no way to enumerate WhatsApp's bot-side block list — `blocked_users` is the one moderation state with no `list_*` counterpart


swampratnz:
## Problem statement

WhatsApp's bot-side block list (`block_user`/`unblock_user`, issue #572,
`docs/SECURITY.md` §18) has no read path. `src/storage/repository/moderation.ts`
exports exactly `blockUser`, `unblockUser`, and `isUserBlocked` — there is no
`listBlockedUsers`, and no tool or digest line references `blocked_users` at
all (verified by grep across `src/`, and confirmed absent from
`adminDigest.ts`, `docs/ARCHITECTURE.md`, and `docs/SECURITY.md`).

This is the exact "a bare count/log is not a 'who' answer" problem the repo
already solved for Discord mutes: `list_muted_members` (issue #487) exists
specifically because the digest's `🔇 N member(s) currently muted` count
"was never meant to provide [who] on its own" (its own docstring in
`src/agent/tools.ts:5134`). `blocked_users` is the newer, simpler,
guild-wide-by-design analog for WhatsApp — it has no `conversation_id` at
all, `PRIMARY KEY (platform, external_id)` — yet it never got the same
enumeration tool.

The only theoretical fallback, `moderation_history`, doesn't cover this: it's
explicitly conversation-scoped (`callerScope()`), shows a raw chronological
action log rather than current state, and never reconciles a `block_user`
row against a later `unblock_user` row. An admin who wants to know "who is
currently blocked on WhatsApp right now" has no way to answer that other
than mentally diffing block/unblock audit rows one at a time, and may not
even see all of them if any fall outside their own conversation scope.

**Who this helps:** admins running a WhatsApp community, who today can block
an abusive or spam sender (`block_user`) but have no way to review or audit
the resulting block list afterward — e.g. to check whether an old block is
still needed, or to confirm a block actually landed.

## Proposed approach

Mirror `list_muted_members` (#487) exactly, for `blocked_users` instead of
`member_warnings`:

1. Add `listBlockedUsers(platform: string, limit = 50)` to
   `src/storage/repository/moderation.ts`, next to `blockUser`/`unblockUser`/
   `isUserBlocked`:
   ```sql
   SELECT external_id, blocked_by, reason, blocked_at
     FROM blocked_users
    WHERE platform = $1
    ORDER BY blocked_at DESC
    LIMIT $2
   ```
   Bound parameters only, same posture as every other query in this file.
[...]

Issue #916, opened 2026-07-31
WhatsApp's `!projects` text command still has no `mine` recall filter — #871's rejection reason (unmerged #869) no longer holds


swampratnz:
## Problem statement

**Who it helps:** WhatsApp members managing their own showcased projects. `share_project`'s edit/remove flow is upsert-by-name, so a member who has shared more than one project (`MEMBER_PROJECT_CAP = 3`) or shared one weeks ago has to recall its exact stored name before they can edit or remove it. `list_projects` (tool) and Discord's `/projects` both gained a `mine: boolean` recall filter for exactly this reason (#867, PR #869). WhatsApp's `!projects` text command (#859/#861) never did.

**Evidence, verified against current `main` (commit `cf31ba1` + `0ffed27`):**

- `src/storage/repository/memberProjects.ts:201` — `listOwnProjects(platform, userId)` exists and is exported (added by PR #869, `git log` confirms `cf31ba1 Add mine filter to list_projects so members can recall their own project names (#869)` is merged to `main`).
- `src/agent/tools.ts` imports `listOwnProjects` and wires it into `list_projects`'s `mine` argument (`tools.ts:4642-4658`), with the empty-state string `"You haven't shared any projects yet."`.
- `src/platforms/discord/slashCommands.ts` has the matching `/projects mine:true` option (confirmed present alongside the tool wiring).
- `src/router.ts:1840-1852`'s `tryWhatsAppTextCommand` still implements `!projects [query]` with exactly two branches — a `query`-driven `searchProjectsFn` call, or (no query) `listRecentProjectsFn`. No third, caller-scoped branch exists; grepped `listOwnProjects` usage across `src/` — only `memberProjects.ts` (definition), `tools.ts`, and `slashCommands.ts` reference it. `router.ts` is absent.

## Why this isn't a duplicate of #871 (rejected)

[...]

Issue #106, opened 2026-07-04
SECURITY review: knowledge_search ignores entry scope — conversation/platform-scoped knowledge leaks to every tier, everywhere


swampratnz:
Finding from a repo-wide security review (owner-requested, 2026-07-04). **Verified against `main`.**

## Problem

The `knowledge` schema, `save_knowledge`, and `list_knowledge` all model a `scope` (`'global'` | platform | conversation id) — but **retrieval ignores it entirely**:

- `searchKnowledge()` (`src/storage/repository.ts`) selects over the whole table with no scope predicate.
- The `knowledge_search` tool (`src/agent/tools.ts`) passes no caller context: `searchKnowledge(args.query)`.

So an admin who saves an entry with `scope: '<private-channel-id>'` — the API shape actively invites this — gets no scoping at read time: **any member (or open-mode guest), in any conversation, on either platform, retrieves it** via semantic search. The mismatch between what the write side promises and what the read side enforces is the security bug; scope is currently decorative metadata. (Also true of the #93 near-duplicate check and #95 relevance-floor path, which compare across all scopes.)

Blast radius is bounded — knowledge is admin-authored, not member PII — but "admin believed it was channel-scoped, bot recites it in a public channel" is a real disclosure class, and SECURITY.md/ARCHITECTURE.md nowhere document that scope is unenforced.

## Proposed fix (one PR)

1. `searchKnowledge(query, caller: { platform, conversationId }, topK?)`: filter to `scope IN ('global', $platform, $conversationId)`. Same filter in the duplicate-check lookup so a private-channel entry can't be "nudged" into view from elsewhere.
2. `knowledge_search` tool passes `caller.platform`/`caller.conversationId` (already in scope in `buildToolServer`).
3. `list_knowledge` (admin browse) keeps its existing explicit-scope filter behaviour — admins curating need to see everything; document that read path as admin-tier in SECURITY.md.
4. Docs: one paragraph in ARCHITECTURE.md ("Memory & learning") and SECURITY.md defining scope semantics as **enforced** from now on.

## Acceptance criteria

- `SECURITY:` test: an entry saved with a conversation scope is returned by `knowledge_search` only in that conversation; a different conversation (same platform) and the other platform get no hit. A `'global'` entry is returned everywhere. Pinned with hand-crafted embeddings (no model download), same technique as the existing repository tests.
[...]

PR #25, opened 2026-07-03
Add proactive super-admin usage alert for shared Max-pool budget

swampratnz:
Closes #22

## Summary

Adds an opt-in proactive usage alert so super admins learn the shared Max
subscription pool is trending toward exhaustion without having to remember
to invoke `usage_stats` manually.

- `src/config.ts` — `USAGE_ALERT_DAILY_REPLIES` (optional int, default
  0/unset = disabled, matching the existing opt-in pattern for
  `INTERACTION_RETENTION_DAYS`/`HEALTH_PORT`).
- `src/usageAlert.ts` (new) — `stepUsageAlertTracker`/`initialUsageAlertTracker`:
  a pure rolling-window latch (mirrors `healthState.ts`'s disconnect
  tracker so it's unit-testable without timers/DB) — `crossed` is set true
  the first tick `outbound >= threshold`, and only clears once a later tick
  sees `outbound` drop back below. `startUsageAlert(adapters)` wires an
  hourly `usageStats(1)` check to it; returns `null` (no timer) when the
  threshold is unset, so a disabled deployment runs zero extra queries.
  On trip: logs at `warn` and DMs super admins via the same
  `sendDirectMessage` path `health.ts`'s disconnect alert already uses.
- `src/index.ts` — wires `startUsageAlert` alongside `startRetentionPurge`/
  `startDisconnectAlerts`, including it in the graceful-shutdown
  `clearInterval` block.
- `docs/ARCHITECTURE.md` — new "Usage & shared Max-pool alerting" section.
- `docs/SECURITY.md` — note under "Abuse / cost runaway".
- `.env.example` — documents the new var.

The trigger is the rolling-24h **outbound reply count**, not `cost_usd`:
SECURITY.md already documents that cost recording can silently degrade
open, whereas reply count is a plain, always-populated row count. Reply
count is still a coarse proxy for actual token/pool draw (short vs long
replies draw differently) — the config comment and docs say so, and
`cost_usd` is surfaced in the alert body as supplementary context when
non-zero. No auto-`pause_bot` — this is alert-only; a super admin decides.

## Security impact

- **No new data exposure** — reuses `usageStats()`, already `super_admin`-only
  via the existing `usage_stats` tool; this just pushes the same aggregate
  counts/cost to the same audience (super admins) proactively instead of
  on-demand. No message content, no per-user identifying detail.
- **No new privileged tool, no new RBAC surface** — reuses the existing
  `sendDirectMessage` super-admin alert path from `health.ts`; no caller,
[...]

PR #929, opened 2026-07-31
Add projects: shared team memory scoped by membership and surface

### LINKED ISSUE #927: Projects: person-scoped shared memory that follows a team across Discord and WhatsApp — `knowledge.scope` already has the slot, `searchKnowledge` can't see who is asking
## Problem

The named consumer is running an **Impact Lab**: a standing team that needs Dave to remember decisions, notes and action items, where the team spans a Discord channel *and* a WhatsApp group, and where only that team may see the content.

Nothing in the tree supports this. `member_projects` is a *showcase* — name, description, link, one row per member, no ACL, no stored content. It is a directory entry, not a workspace.

The two obvious cheap answers both fail:

- **"Use a private channel as the boundary."** Works while a project lives in one conversation. A team spanning Discord *and* WhatsApp is two conversations in two identity namespaces, so `conversation_id` cannot be the membership key.
- **"Key it on the member."** `searchKnowledge(query, caller: { platform, conversationId }, topK)` **does not receive a user id at all** — only where the question was asked. Person-keyed access control is not expressible without threading identity through that call. This is the single biggest implementation surprise in the change and is called out again under Scope.

## What already exists (do not rebuild)

- **`knowledge.scope`** is already a free-text discriminator — `'global' | <platform> | <conversationId>`, no CHECK constraint. A project is a fourth value, `project:<id>`. No new storage engine, and embeddings, near-duplicate detection on save, retrieval counts, freshness/staleness signals and `merge_knowledge` all come along for free.
- **`persons`** already groups the `community_users` rows that are the same human across platforms, created only via the admin-tier `link_member` tool, "never inferred from message content". Project membership keys on `person_id` and inherits that.
- **`link_member`'s tool shape** is the template for the membership admin tool: confirmation-gated, explicit "NEVER changes anyone's tier", a cross-platform authority gate, and a refusal for env-configured super admins.

## The invariant this must not break (issue #106, verbatim precedent)

`searchKnowledge`'s own doc comment records the exact failure this proposal risks reintroducing:

> **SECURITY: issue #106** — `scope` used to be write-only metadata; an admin who saved a conversation-scoped entry had it recite to every tier, everywhere.

A `project:<id>` scope that is written but not enforced at retrieval **is issue #106 again, with a private team's content instead of one conversation's.** Enforcement belongs in SQL in `searchKnowledge`, not in prompt text, and that is where the `SECURITY:` tests go.

Related and equally binding: `opts.scopeRestriction: 'global-only'` (issue #165) exists so a gated guest is never served a platform- or conversation-scoped entry. Project entries must fall on the same side of that gate — a guest must never see project content under any code path.

## Design decisions already settled

Recorded so they are not re-litigated during the build.

**1. Membership is by person; rendering is by conversation. Both are checked.**
Person-keyed scope alone means project content becomes retrievable in *any* conversation that person is in — a Lab member asks about the venue in `#general` and Dave answers with private content in front of the whole server. So a project has a member set **and** a set of bound surfaces, and a project entry surfaces only when the asker is a member **and** the current conversation is bound (or is a DM to a member). The channel stops being the membership boundary and becomes the rendering boundary.

**2. A project grants data scope, never a tier.**
`persons` already codifies this — the schema comment says linking "never touches `role` (tier stays per-platform-row by design)". Project membership does the same: it narrows which data a tool sees, never which tools exist.

**3. Project-scoped *tools* are out of scope here, and when they come, they gate inside the tool body.**
`core.ts` documents `allowedTools` as derived from role and "never from message content, and **only ever a subtractive filter**". Making the surface project-aware would convert a subtractive filter into an additive one, so a bug could conjure a tool rather than only drop one. The eventual design is a member-tier tool that asserts project membership + capability + surface binding in its own body, the way privileged tools already re-assert their tier. Not in this slice.

**4. `forget_me` keeps the decision and drops the person** — see the dedicated section below.

## Scope (one PR)

**Schema** — four small tables plus one scope convention:

- `projects` — id, slug, name, `brief`, created_by, created_at, archived_at
- `project_members` — project_id, person_id, added_by, added_at
- `project_surfaces` — project_id, platform, conversation_id
- `project_actions` — id, project_id, title, owner_person_id, due_at, status, created_at
- Notes/decisions/facts are `knowledge` rows at `scope = 'project:<id>'` — no new content table

The dividing line: **if you would find it by asking a fuzzy question it is `knowledge`; if you would filter or sort it, it is a table.** `project_actions` needs update-in-place and status transitions, which `knowledge` does not model.

**The retrieval change** — `searchKnowledge` currently resolves global + caller platform + caller conversation. It must additionally resolve project entries for projects the asker belongs to, intersected with surfaces the current conversation is bound to. **Its `caller` parameter must gain the asker's identity**, since `{ platform, conversationId }` cannot resolve a `person_id`. Every call site needs updating; this is the bulk of the diff and all of the risk.

**Admin tools** — create/archive a project, add/remove a member, bind/unbind a surface. Modelled on `link_member`: admin tier, confirmation-gated, never role-touching, super admins not manageable here.

**Explicitly NOT in this slice:** project-scoped tool access, any external egress (Google Docs/Drive), file or attachment storage, text extraction from uploads, and email as a surface. Documents are represented as *references only* — a URL stored as text and never fetched, exactly as `member_projects.link` already is.

## Security / privacy impact

- **New authorization axis**, which is the main thing to review. Mitigated by keying on `persons` (admin-set, never content-derived), enforcing in SQL, and requiring both checks from decision 1.
- **No new tier, no new tool-surface derivation, no egress, no stored bytes.** The three postures most likely to be eroded by a "workspace" feature are all held.
- **`brief` is data, not authority.** It is member-adjacent text reaching the prompt. `systemPrompt.ts` already handles this tension for personas — voice rules never override the security section — and a project brief needs the same framing: context only, never instructions, never anything reading as a permission grant.
- **Guests**: project content must be unreachable under `scopeRestriction: 'global-only'` (#165).

## `forget_me`: keep the decision, drop the person (owner decision)

`purgeSingleIdentity` currently runs `DELETE FROM knowledge WHERE source_user_id = $1`, so as things stand a departing Lab member's `forget_me` would hard-delete every project decision they authored — gutting the project's shared history as a side effect of an unrelated privacy action.

**Decision (owner):** for `project:`-scoped entries, **null the author link and keep the row.** Precedent: `knowledge_candidates` "only nulls the link for accepted/declined ones".

Three implementation constraints that follow, and that a build worker must not soften:

1. **Narrow the exception to `project:` scope only.** Global-, platform- and conversation-scoped `knowledge` authored by the member must still hard-delete exactly as today. This is an exception for shared project memory, not a general weakening of erasure. A `SECURITY:` test should pin both halves — project entry survives with a null author, non-project entry is gone.
2. **Nulling the author is not the same as scrubbing the content.** A decision can name or describe the person in its text ("Chris is hosting at his place"). Nulling `source_user_id` removes the authorship link, not personal information *inside* the entry. This is a real residual: the erasure is partial by design. It must be stated plainly in `docs/SECURITY.md` and in whatever `forget_me` tells the member, so nobody is told their data is gone when some of it is retained.
3. **Companion rows follow the same rule.** `project_members` hard-deletes (it is pure identity). `project_actions.owner_person_id` nulls, so the action survives unowned. `projects.created_by` nulls.

## Acceptance criteria

- [ ] A project spanning a Discord channel and a WhatsApp group shares one memory; both platforms' members reach it.
- [ ] **SECURITY:** a non-member never retrieves project content, on any platform, in any conversation, via `knowledge_search` or any other recall path.
- [ ] **SECURITY:** a *member* asking in a conversation **not** bound to the project does not receive project content (decision 1's second check).
- [ ] **SECURITY:** a gated guest never receives project content, including under `scopeRestriction: 'global-only'` (#165).
- [ ] **SECURITY:** project membership is settable only by the admin tool, never from message content; a message claiming membership changes nothing.
- [ ] Adding someone to a project changes no tool surface and no tier — asserted directly against `toolsForRole()`.
- [ ] **SECURITY:** `forget_me` nulls the author on `project:`-scoped entries and keeps them, hard-deletes the member's non-project `knowledge` rows, hard-deletes `project_members`, and nulls `project_actions.owner_person_id` / `projects.created_by`. All five asserted.
- [ ] `forget_me`'s member-facing wording and `docs/SECURITY.md` both state that project entries are retained with the author unlinked, and that content naming the member is not scrubbed.
- [ ] Existing `searchKnowledge` behaviour for global / platform / conversation scopes is unchanged — same results, same ordering — with the widened signature.
- [ ] `npm run context:check` passes with the new modules described in `docs/agents/module-map.md`.

## Metric

**Participation**, read at project level: the Lab's decisions and actions being retrievable later without a human re-reading the channel scrollback. Concretely, project-scoped entries accumulating and showing non-zero `retrieval_count` — the same signal `knowledge` already tracks — rather than the memory being write-only.

## Provenance

Design worked through with the repo owner in session, from the Impact Lab use case. The rejected #925 is adjacent but distinct: that proposed generalising async job completion, this proposes scoped storage. Nothing here depends on it.

swampratnz:
## Summary

Closes #927.

A **project** is a standing team's shared memory — the Impact Lab use case — that follows the *team* across Discord and WhatsApp instead of living in one channel. An admin creates a project, adds members, and binds the conversations it may be discussed in; members then record decisions with `project_note` and find them later with `project_recall`, from either platform.

**The access model is two checks, both in SQL in `visibleProjectIds`, never re-derived by callers:**

- **Membership (who)** — the caller's own platform identity, or any identity sharing their `person_id`, so one human reaches the project from both platforms once `link_member` has linked them. Keyed on the platform identity rather than `person_id` because `linkMembers` *merges* person rows (keeps the lower id, drops the other), which would otherwise need repointing on every link.
- **Surface (where)** — an explicit `project_surfaces` binding for this conversation, or a DM.

Membership alone is deliberately not sufficient. Without the second check a member asking in a public channel would have private notes recited in front of everyone — issue #106's failure mode with a team's content instead of one conversation's.

**A project grants data scope only, never a tier.** `project_recall` / `project_note` / `project_list` sit in `MEMBER_TOOLS` for *every* member and are simply inert without a visible project, so the per-turn tool surface stays tier-derived and **only ever subtractively filtered** — a bug in project logic cannot conjure a tool. Pinned by a test comparing `toolsForRole` output either side of a membership grant. Same rule `persons` already states for identity linking. `project_bind_here` takes no conversation-id argument, so nothing can bind a channel the admin isn't actually in.

### One deliberate departure from the issue's written plan

#927 specified storing notes as `knowledge` rows at a `project:<id>` scope value. **I changed this during implementation** after auditing the readers: `knowledge` has ~20 that are unrestricted by default (`listKnowledge`, the duplicate/conflict pair-finders, the link-rot checker, the staleness readers, every get-entry-by-id path). Private project content there would have been one un-audited caller away from an admin-facing view, and every future reader would be a new leak site — a bad trade for a security boundary.

`project_notes` has its own embedding and its own readers, all project-aware by construction. It also makes the `forget_me` rule local instead of an exception carved into the global `DELETE FROM knowledge`. **Cost:** project recall is a separate call from `knowledge_search` rather than merged into it, so `/kb` and the zero-cost knowledge shortcut don't reach project content. Noted as a follow-up.

## Security / privacy impact

This adds a **new authorization axis**, so it is the main thing to review. `docs/SECURITY.md` gains a full section (#25).

- Membership and surface bindings are set by admin tools only, **never from message content** — same rule as roles.
- Notes re-enter the model's context on recall and are quarantined with `untrusted()`, as `community_digest` and `admin_digest` already are. A project `brief` is context, never authority.
- `reference_url` is stored verbatim and **never fetched** (`member_projects.link`'s rule). No file storage, no external egress, no new outward credential.
- Archiving is a revocation, not a label — `visibleProjectIds` excludes archived projects.

**`forget_me` is partial here, per the owner decision on #927, and it is the only such exception in the codebase:** `project_members` hard-deletes (pure identity; access ends immediately) while `project_notes` rows are kept with authorship nulled, so a departing member's erasure cannot silently gut a standing team's decisions. The exception is scoped to project content **only** — `DELETE FROM knowledge` is untouched and ordinary member-authored knowledge still disappears entirely, pinned by its own test.

**Documented residual (NZ Privacy Act 2020):** nulling authorship removes the *link*, not personal information the note's own text may contain. The erasure is partial by design; `docs/SECURITY.md` states this. One acceptance criterion from #927 is **deliberately not implemented here** — making `forget_me`'s own reply say so is member-facing copy the repo owner should word, and is called out in the issue rather than guessed at.

**Not in this slice**, and named as such in the docs: project-scoped tool access, external egress (Google Docs/Drive), file storage or text extraction, and email as a surface.

## How verified

Against a local **Postgres 16 + pgvector 0.6.0**, installed and run for this change rather than deferred to CI. All figures below are **after** the `main` merge in `defd184`, each on a **freshly created database**:

- `npm run typecheck` (incl. `typecheck:tests`) — clean. `tests/projectScope.test.ts` is added to `tsconfig.tests.json`'s ratchet, so it can never regress.
- `npm run lint`, `npm run format:check` — clean.
- `npm run migrate` — applied, then **run a second time** to confirm the new tables and trigger are replay-safe (the repo's stacked-DDL wedge class), and **re-run after merging `main`** per CLAUDE.md.
- `npm test` — **3235 pass, 0 fail**.
- `npm run test:security` — manifest **1214 across 153 files**, exact match.
- `npm run build`, `npm run context:check` — clean.

**10 new `SECURITY:` tests** in `tests/projectScope.test.ts`, including two positive controls — without them every negative assertion would pass vacuously against a `searchProjectNotes` that just returned `[]`. They cover: non-member in a bound channel gets nothing; member in an *unbound* conversation gets nothing; DM works without a binding; a linked WhatsApp identity gains access and an unlinked one does not; writes are gated by the same two checks; archived projects are invisible; `forget_me` keeps the note and drops the membership; ordinary knowledge is still hard-deleted; and membership changes no tool surface.

### Merge notes (`defd184`)

Two conflicts, both from concurrent work on `main`:

- **CHANGELOG.md** — both sides prepended to today's `### Added`. Kept both, main's (#916) first so this reads as a pure insertion.
- **tests/tools.test.ts** — both sides bumped the *same* super-admin `community_info` char cap for different reasons (#924's `list_blocked_members`, #927's project lines). Kept both attributions and raised the cap to clear the merged total.

Worth flagging for review: the **admin** cap auto-merged to this branch's value and then no longer cleared #924's added clause (measured 4100 against a 4080 cap). Git resolved that line silently — an auto-merge keeping one side's number is not the same as a number still being correct after the merge. It is now 4130 with #924 credited alongside #927.

### On flaky-looking failures

Intermediate runs showed unrelated failures in `backgroundJobCostAlert`, `usageStats` and an `admin_digest` zero-signal test. All pass on a fresh database, and both a full-suite run on **clean `main`** and `main`'s own `tools.test.ts` in isolation reproduce failures of the same class — so this is the shared-database cross-file contamination CLAUDE.md documents, not something this diff introduces.


swampratnz:
Thanks — the review found a real gap. Pushed `eafbb31`. One half fixed, one half I'm pushing back on with justification, as the review itself offered.

## Fixed: no revoke tool

You're right, and it was a genuine defect independent of the confirm question. `removeProjectMember` existed in the repository but was never wired into `buildToolServer`, so a mistaken grant could only be undone by archiving the whole project. Shipping an access-control feature with a grant and no revoke isn't defensible.

`project_remove_member` is now exposed — admin tier, audited, same `resolveMemberTarget` cross-platform gate as its sibling — with a `SECURITY:` test that runs a positive control first, then asserts removal revokes **both** reads and writes immediately.

That test also pins a distinction worth having explicit, because the two removal paths are easy to conflate:

- **`project_remove_member`** deletes the membership row and leaves the member's notes with the project, authorship intact.
- **`forget_me`** keeps the note and *nulls* the author.

Opposite operations, both now pinned.

## Not changed: the CONFIRM gate — the analogy doesn't hold

The review reads `project_add_member` as modelled on `link_member`. The correct precedent is **`add_member`**.

This repo's CONFIRM gate is for **destructive or irreversible** actions: `delete_knowledge`, `remove_member`, `unlink_member`, `grant_admin`. `link_member` is gated for exactly that reason, and its own description says so — linking permanently expands what a single `forget_me` **erases**, across both identities. That is an irreversibility gate, not an access-expansion gate.

Granting project access destroys nothing, and as of this push it is undone in one call. Meanwhile `add_member` — which grants access to the entire bot, a strictly *larger* grant than one project's notes — is admin-tier and audited with **no** confirm. Gating this one and not that would make a subset stricter than its superset.

The review asked for a confirm step *"or explicit written justification for why this action is lower-consequence than `link_member`'s"*. This is that justification, and I've recorded it where it'll be found again rather than only here: at the call site in `tools.ts`, and in `docs/SECURITY.md`'s section 25.

Happy to add the gate anyway if a maintainer prefers belt-and-braces — it's a two-line change — but I don't think consistency with the existing convention supports it.

## Verification

Fresh database per run, local Postgres 16 + pgvector: typecheck (incl. tests), lint, format:check, migrate, `npm test` (**3236 pass, 0 fail**), `test:security` (manifest **1215 across 153 files**, exact match), build, context:check — all green.

One note for whoever reads the CI result: full-suite runs against a *reused* database intermittently fail a couple of `usageStats` / `backgroundJobCostAlert` / `admin_digest` tests. Those reproduce on clean `main` too and pass on a fresh database — the cross-file contamination CLAUDE.md documents, not this diff. Every figure above is from a freshly created database.

---

swampratnz:
All four addressed in `10a64ce`. This review was right on every point, and the first was a real leak rather than a hardening nit — thank you.

The three findings share one root cause worth naming: **`visibleProjectIds` deliberately checks only `project_members` and never tier.** That's the correct shape for a data scope, but it means tier has to be enforced *around* it — and in three places it wasn't.

**1. `remove_member` didn't cascade to `project_members`** — confirmed and fixed. `removeMember` only `DELETE`s the `community_users` row, and `project_members` has no FK to it (keyed on the platform identity so visibility survives the person-row merges `linkMembers` performs), so nothing cascaded. Gated deployments stop the identity at the router, but in **open mode** `toolsForRole` hands a guest the full `MEMBER_TOOLS` surface — so a member an admin had just removed kept read *and* write access to the team's private notes. Now deleted explicitly, mirroring what `purgeSingleIdentity` already did for the same stated reason.

**2. Missing open-mode-guest re-check** — fixed. `project_recall` / `project_note` / `project_list` now `assertAtLeast(caller.role, 'member', …)` in the handler, matching the convention `share_project` / `set_my_interests` / `who_is_into` / `find_helper` / `community_digest` follow and that `rbac.ts` documents. Also noted on the `MEMBER_TOOLS` entries so the next reader sees it.

**3. `project_add_member` accepted an arbitrary identity** — fixed. It now requires the target to be a known community member, exactly as `link_member` does, and points at `add_member` otherwise. You were right that this is the enabler for the other two: without it a membership row can exist for an identity that never passed `add_member`.

**4. Unreachable `archiveProject` / `unbindProjectSurface`** — fixed rather than documented around. You were right that `SECURITY.md` and `ARCHITECTURE.md` described archiving as an operative revocation while nothing could invoke it. `project_archive` and `project_unbind_here` are now admin tools, so the documented lifecycle is real.

**Tests:** two new `SECURITY:` tests, each with a **positive control first** so neither can pass vacuously — `remove_member` revokes access as a matter of *storage* (asserting the row is gone, not merely unreachable), and all three member tools reject a guest-tier caller before any DB access.

On the CONFIRM point from the previous round: that was settled as well-justified in this review, so I've left it as is.

**Verification** — fresh database per run, local Postgres 16 + pgvector: typecheck (incl. tests), lint, format:check, migrate, `npm test` (**3238 pass, 0 fail**), `test:security` (manifest **1217 across 153 files**, exact match), build, `context:check` — all green.

---

swampratnz:
All four addressed in `c48e0cf`. Every point was valid, and they share a theme worth naming: the admin surface is the part of this feature most directly on the security spine, and it was the one place I skipped the repo's own conventions.

**1. Untested tier checks on the admin tools** — fixed. Handler-level `Permission denied` tests now cover every project-management tool against both guest and member tier, matching how the repo pins `grant_admin` / `link_member` / `remove_member` and ~50 others.

**2. Untested known-member guard** — fixed, and you were right that this one stung: `docs/SECURITY.md` calls it "layer 3" and says it closes an open-mode leak, so shipping the fix without a test was the weakest spot in the diff. It now has one, with a **positive control** — the same call succeeds once the target *is* a member, so the refusal is the guard firing rather than an unrelated error.

**3. `rbac.test.ts` not updated** — fixed, and pinned in **both** directions. The admin tools must be in `ADMIN_TOOLS` and unreachable from member/guest; the three member tools must stay in `MEMBER_TOOLS` and **not** become admin-gated. The second half matters as much as the first: the whole design depends on project membership never widening a surface, so drift either way is now a red test.

**4. Unused exports** — wired rather than dropped. You're right that an admin could create, add, remove, bind, unbind and archive but had no way through the bot to see what exists, who's in it, or what it's bound to. `project_info` (admin, read-only) covers all three exports and returns membership and binding **metadata only** — note content stays reachable solely through the two-check member path.

Worth noting this is the same class of finding as last round's unreachable `archiveProject`: exports with no caller, and code paths with no test, both read as "someone will wire this later" and neither survives a refactor. Two rounds in a row, so it's a pattern in how I built this, not a one-off.

**Verification** — fresh database per run, local Postgres 16 + pgvector: typecheck (incl. tests), lint, format:check, migrate, `npm test` (**3242 pass, 0 fail**), `test:security` (manifest **1221 across 153 files**, exact match), build, `context:check` — all green.

One transparency note: an intermediate full-suite run on a **reused** database failed three `usageStats` delta-reconciliation tests. They passed on a fresh database in the same session with no code change, and clean `main` reproduces the same class. If CI shows those, re-run before treating them as real.

---

swampratnz:
Addressed all three points in 1285ec2 rather than leaving them for a maintainer to adjudicate.

**1. `project_archive` was a one-way door — fixed, not just documented.** The review is right that this is the odd one out. Archiving revokes read access for everyone *including the project's own members*, and nothing on the tool surface could give it back. That leaves two honest options: CONFIRM-gate it, or make it reversible. Reversible is the better answer, and it makes the whole surface consistent — every project revocation now has a matching restore that touches nothing else:

| revoke | restore |
| --- | --- |
| `project_remove_member` | `project_add_member` |
| `project_unbind_here` | `project_bind_here` |
| `project_archive` | `project_unarchive` ← new |

`unarchiveProject` only clears `archived_at`. Archiving never touched membership or surface rows, so the restore returns *exactly* the access that existed before and grants nothing new. That claim is pinned by a `SECURITY:` test asserting both halves — the member can read again **and** a non-member still cannot — plus a no-op test (an active project and an unknown slug both report no change, and no row is conjured) and a tool-level round trip covering the reply wiring. `docs/SECURITY.md` now states the rule going forward: if a future change removes a restore path, that revocation has to become CONFIRM-gated in the same diff.

**2. `project_info`'s guild-wide scope — intentional, and now written down.** It was deliberate, but the reasoning existed only in my head, which is the actual defect. The "admin data access is scoped in SQL to conversations the admin is in" rule governs *member content* — messages, notes, things said in confidence. `project_info` returns only the administrative register: names, slugs, who holds access, which conversations are bound, and never a single project note. An admin who could only administer projects they happened to belong to could not audit the grants they are accountable for, and could give themselves that visibility with one `project_add_member` call anyway — so the narrower scope would cost the audit trail without buying any confidentiality. Same precedent as `list_roster`/`blocked_users`, as you noted. Reasoning now sits at the call site and in `docs/SECURITY.md`; behaviour unchanged.

**3. The `project_create` TOCTOU — fixed.** Now `ON CONFLICT (slug) DO NOTHING`, matching `addProjectMember`/`bindProjectSurface` in the same file. `createProject` returns `Project | null` and the tool renders the friendly "already exists" from the null. The test issues both duplicate creates *concurrently* and asserts the raw Postgres unique-violation text never reaches the reply, so a check-then-act implementation can't pass it by winning the race.

Verified against a local Postgres 16 + pgvector: `typecheck`, `lint`, `format:check`, `migrate`, `npm test`, `test:security` (manifest regenerated, `projectScope.test.ts` 12 → 14), `context:check`, `build`.

One pre-existing failure to flag, unrelated to this diff: `admin_digest: returns the fixed 'Nothing to report right now.' text when every signal is zero` fails locally against a freshly migrated DB **on the pre-change tree as well** (I stashed and re-ran to confirm). It's within-file fixture bleed in `tools.test.ts`, not something this change introduced.

---

swampratnz:
Round 5 was right on every point. All five addressed in a115218.

**1. `project_note` had no length cap — valid, and the worst of the five.** I checked the claim rather than taking it: every other member-writable free-text field in `tools.ts` is capped, and `project_notes` was the one new member-writable table with no bound at all. Fixed with the same defence-in-depth `createKnowledgeTip` uses — zod `.max()` at the tool layer on `content`/`title`/`reference_url`, **plus** a `slice()` inside `saveProjectNote`, because that's an exported repository entry point a later caller could reach without going through the tool schema.

One thing the review didn't ask for but the numbers demanded: the caps are chosen so `title\ncontent` always fits inside `embed()`'s own `slice(0, 4000)`. A note whose stored text outran its embedding would be silently unfindable *by its own tail* — worse than refusing the write. The test pins the relationship between the constants, not the constants themselves, so it survives someone retuning them.

**On the "arguably a daily-write cap" — I've added one, because the repo's own closest precedent settles the argument.** My first instinct was that daily caps here guard *outbound social* actions (`find_helper`, `request_project_connection` both notify a human). But `createKnowledgeTip` is the real analogue — a member writing free text into a table — and it carries a DB-backed rolling-24h cap counted *inside* the `INSERT`, so it can't be raced by concurrent turns and survives a restart. `saveProjectNote` now does the same.

I deliberately did **not** copy the 3/day those caps use. Every existing cap guards an action that costs a *human* something — an admin review-queue entry, a DM to another member — so 3 is a courtesy budget. A project note costs only storage, inside a team an admin has already trusted the member into, and a team minuting a meeting will legitimately record many in one sitting. So this is an abuse ceiling, not a usage budget: high enough no honest team meets it, low enough that one member can't bloat the table unattended. It's also **per-member, not per-project**, with a test for that specifically — a per-project cap would let one abuser deny service to their entire team.

**2. `project_create`'s `name`/`brief`** — capped on the same principle. Named `TEAM_PROJECT_*` rather than `PROJECT_*` because `memberProjects.ts` already owns `PROJECT_NAME_MAX_CHARS` for the unrelated public showcase row, and colliding those two would be a genuine trap.

**3. `projectsVisibleTo` in `schema.sql:1142`** — correct, that function has never existed; the real name is `visibleProjectIds`. Fixed, and I grepped for other stale references (none).

**4. Tool-handler coverage for `project_bind_here` / `project_unbind_here` / `project_remove_member`** — added, covering the success *and* the no-op reply on each ("was already bound", "Not a member of"), matching what their siblings already had.

**5. Archived-project edits reading as a silent no-op** — agreed, and worth more than surfacing. `getProjectBySlug` not excluding archived projects is deliberate (an admin must be able to fix membership before an unarchive), but since `visibleProjectIds` excludes them from every read and write, the admin's edit does nothing observable until then. Every membership and surface reply now carries an explicit ARCHIVED warning naming `project_unarchive`, with a test asserting an *active* project doesn't get the warning so it can't degrade into noise.

Verified on a **freshly created** database (Postgres 16 + pgvector): `typecheck`, `lint`, `format:check`, `migrate`, `npm test` — **3252 pass, 0 fail** — `test:security` (manifest regenerated: `projectScope.test.ts` 14 → 16, **1225 across 153 files**), `context:check`, `build`.

Note on the `admin_digest` flake I mentioned last round: it does **not** appear in a full-suite run on a fresh database. It only reproduces when `tools.test.ts` runs alone, and it does so identically on the pre-change tree — so it's within-file fixture ordering, and this diff is clean either way.

Still a draft, awaiting the repo owner's decision.

---

swampratnz:
Thanks — LGTM noted, and I've acted on the non-blocking observation in 78139c1 because it turned out to matter more than "non-blocking" suggested.

The review is right that `docs/SECURITY.md` §25 asserts a `must` the tree doesn't meet. The deferral itself is fine and deliberate — that text is copy about a privacy guarantee, so it's the repo owner's to word, not mine to guess. The defect is that §25 stated the requirement *alongside* the implemented ones, so it read as satisfied. In a security document that's the thing worth fixing, and it's my wording that did it. §25 now says plainly that the member-facing half is outstanding, and why.

**The part I'd flag beyond the review's note:** the review suggested tracking it so the `must` doesn't sit unmet indefinitely. Tracking it *in #927* wouldn't have worked — this PR says `Closes #927`, so that unchecked acceptance criterion would have been closed unmet and vanished with the issue. So I've filed it as **#930** (owner-worded copy, no schema or behaviour change) and pointed §25 at that instead. Left unlabelled beyond `proposal`/`theme:knowledge` so it sits for your triage rather than entering the build lane.

#930 also records one open question I don't think a worker should decide: whether `forget_me`'s confirm prompt should *name* the affected projects. Clearer for the member, but it recites project names into a possibly-public conversation at the exact moment someone is asserting a privacy right. My recommendation in the issue is to keep the prompt generic.

Docs-only change: `lint`, `format:check` and `context:check` clean; no code touched, so the round-5 verification (3252 pass / 0 fail on a fresh DB, manifest 1225 across 153 files) stands.

Still a **draft** — over to you to merge. Worth noting it touches `docs/SECURITY.md`, which is a governance path, so it needs a human merge regardless.

---

swampratnz:
Second LGTM noted. One thing in it needed acting on, in cb78fa7 — the concurrency note was right, and I'd claimed the opposite.

**I measured it rather than conceding or dismissing it.** 8 concurrent statements against a cap of 3 inserted **all 8**. So the review's framing understates it: it isn't that "two simultaneous inserts could both read the pre-insert count," it's that the overshoot is bounded by the size of the concurrent burst and not by the cap at all. Each statement takes its own snapshot under READ COMMITTED, so every simultaneous writer sees the same pre-insert count and every one of them lands.

That makes my round-5 wording wrong, and it went into `docs/SECURITY.md` as "counted inside the `INSERT` statement itself so it cannot be raced by two concurrent turns." A security document asserting a guarantee the code doesn't provide is precisely the defect I fixed one commit earlier in 78139c1 — this one was mine, in the same file, in the same review cycle.

Corrected in all three places (the doc, the constant's docblock, the `INSERT`'s comment). The accurate statement: the cap is **restart-proof** — it can't be reset by bouncing the process, which is the actual point of counting in the DB rather than in memory — but it is **not atomic**, and that's tolerable only because it's an abuse ceiling on storage inside a team the member is already trusted into. It is not an authorization check; `visibleProjectIds` is what gates access. The rate-cap test now says it pins the *sequential* guarantee only, so it can't later be misread as proof of atomicity — and I deliberately didn't add a concurrent assertion, since one would be flaky by construction.

**I have not changed the code.** A per-member `pg_advisory_xact_lock` around the count and the insert would make this a hard bound, and it's one line — but it would diverge from the repo-wide `createKnowledgeTip`/`createSuggestion` pattern, and introducing a new concurrency primitive on a PR that's already been signed off twice is your call, not mine. It's recorded in §25 as the known fix if you ever want the stronger guarantee.

Agreed on the rest of your framing: this is inherited from the existing pattern, not introduced here, so it isn't a regression.

Comments and docs only — no code path altered, no `SECURITY:` test added or removed, so `security-floor.json` is unchanged. Verified on a freshly created database: `typecheck`, `lint`, `format:check`, `context:check` clean, `npm test` **3252 pass / 0 fail**. `test:security` and `build` were still running locally when I pushed; CI adjudicates, and I'll act on it if either goes red.

On your closing note — `build` was still in progress at review time on both of the last two heads. It hasn't gone red at any point; I'll confirm green before this is merged. Still a **draft**, and it touches `docs/SECURITY.md`, so it needs a human merge regardless.

---

PR #956, opened 2026-08-02
Agent-base Phase 1: the declarative tool registry — split tools.ts, derive tiers/platforms/flags from one ToolDef table

swampratnz:
## Summary

The highest-leverage item of `docs/AGENT-BASE-PLAN.md` Phase 1 (item 2), built to the reviewed design committed here as `docs/TOOL-REGISTRY-DESIGN.md`. Six commits, each individually gate-green, mirroring the design's strangler sequence — review commit-by-commit.

**What changed.** All 117 tools now live as declarative `ToolDef` entries (`{name, minTier, platforms?, featureFlag?, readOnlyHint, schema, handler(args, ctx)}`) across 28 per-domain files under `src/agent/tools/`, composed into one `TOOL_REGISTRY`. Every handler body moved **byte-verbatim** (verified mechanically at each stage); the only change is the capture mechanism — handlers destructure a `ToolContext` built once per turn by `makeToolContext` (`tools/context.ts`), which owns the security kernel: `audited` (audit row + `'system'`-priority super-admin alert pairing) and `requireConfirm` (the CONFIRM gate's forgeable-pending-notice sanitize) now exist in exactly one file instead of being closure-bound over 5,500 lines.

**What's derived instead of hand-maintained.** `toolsForRole`, the three tier arrays, `DISCORD_ONLY_TOOLS`, the feature-flag filter, and the SDK server's `tools:` array are all derivations of the registry. `rbac.ts` drops from ~440 to 93 lines (keeping its exact export surface — every consumer compiles unchanged); `tools.ts` drops from 8,331 to 231 (a barrel preserving every existing import path — the 21 test files importing from it are untouched); `core.ts`'s `FEATURE_FLAGGED_TOOL_GROUPS` (whose flag booleans froze at import time) is replaced by per-turn predicate evaluation against live config. A new `auth/tiers.ts` leaf (verbatim `atLeast`/`assertAtLeast`/`TIER_ORDER`) breaks the rbac→registry→domains import cycle; `rbac.ts` re-exports them.

**The four-places-per-tool problem is dead.** Adding a confirm-gated admin tool was 11 edits across 6 files with four *silent* failure modes (tier-array omission = dead code in production, which shipped once as `response_latency`). It is now one `defineTool` entry in one domain file — the tier, platform, and flag registrations cannot be missed because they don't exist separately.

**What deliberately did not change** (per the design): tool handler behaviour, `pendingActions.ts`/router CONFIRM flow, the hand-written `community_info` capability prose and its coverage-map tests (rundown generation is the design's deferred PR H), `MODERATION_ACTION_KINDS`, the config schema, and the test-file layout.

## Security / privacy impact

Tool gating is the security spine, so this is the load-bearing section. The tool surface per (role, platform, flags) is **set-identical** to before — enforced by the pre-existing relational tests (`toolsForRole` vs exported arrays, member-turn non-leak, platform exclusion, flag exclusion, the auto-answer `MEMBER_TOOLS` ceiling), all passing unchanged, plus a new `tests/toolRegistry.test.ts` invariant suite (4 `SECURITY:` tests): tier arrays partition the registry exactly; WhatsApp surfaces exclude every `platforms:['discord']` def; member/guest surfaces never contain an admin/super name; the flag filter is provably subtractive and 1:1 with `featureFlag` defs. During the strangle, a cross-check test held the registry and the hand arrays in agreement at every intermediate commit; the flip replaced it with the invariant suite. Handler-side `assertAtLeast` calls are untouched (defence in depth). `requireConfirm`'s sanitize and `audited`'s alert pairing moved verbatim into the kernel — the SECURITY tests covering confirm-flow sanitisation, tool gating, and the feature-flag/config scans all pass (several source-scan tests were repointed to the files the code moved to, assertions unchanged). Process-singleton rate reservers each moved exactly once with their domains (never re-declared, so no cap silently doubled) and stay barrel-re-exported for their adapter importers. Tool ORDER in the SDK listing changed (registry order), which churns the prompt cache once at deploy — a one-time cost, deterministic thereafter.

## How verified

Full suite from a clean checkout of the final tree, independently of the per-stage runs: `npm run typecheck` clean (src + tests ratchet); `npm test` — 3310 tests, 2457 pass, **0 fail** (853 skipped: DB-backed tests without local `DATABASE_URL`, by design); `npm run test:security` — **1244 SECURITY tests, exact manifest match across 159 files** (net +4: the registry invariants, floor regenerated in-diff); `npm run context:check` — 90 sorted entries covering all 41 required paths (28 new module-map entries, 🔒 on `tools/context.ts` and `tools/index.ts`); `npm run format:check`, eslint on all touched files, and `npm run build` clean. Every handler body was mechanically diffed against its pre-move original at each stage (byte-identical). One known cross-file timing flake (`repeatQuestionShortcutRouter`) surfaced once mid-build on an unrelated diff and passes in isolation — not introduced by this work.

https://claude.ai/code/session_01Hkgg61b9V5K1tDwE6t6kUs

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hkgg61b9V5K1tDwE6t6kUs)_


swampratnz:
Re the red `review` check: the automated reviewer hit its 60-turn budget (`error_max_turns` after 61 turns / ~274s) working through this PR's 9,852-insertion diff and produced no final review text, so the workflow's fail-closed guard posted the "no output captured" notice and failed the check — working as designed for a no-verdict run, not a finding against the diff.

This PR is a deliberate exception to normal sizing (the whole registry conversion in one PR, per maintainer request), so the intended review path is human, structured for exactly that:

- **Six commits, each individually gate-green**, mirroring `docs/TOOL-REGISTRY-DESIGN.md`'s strangler plan — review commit-by-commit (design doc → kernel + dev-team/image-gen → member domains → admin domains → super-admin → the flip).
- Every handler body was moved **byte-verbatim** (mechanically diffed per stage), so the reviewable surface is the metadata, the kernel (`tools/context.ts`), and the flip commit's derivations — not 9,800 lines of prose.
- The tool surface per (role, platform, flags) is pinned set-identical by the pre-existing relational tests plus the new `tests/toolRegistry.test.ts` invariant suite (full run: 3310 tests, 0 fail; security floor exact at 1244/159 files).

Note this PR touches `tsconfig.tests.json` (typecheck config), which is a governance path — auto-merge could never take it regardless of verdict, so a human merge was always the endpoint. If a re-run of the automated review is wanted for the record, any push (or re-run of the workflow) will re-trigger it, but it will likely hit the same turn ceiling on a diff this size.

---

PR #960, opened 2026-08-03
Consume @swampratnz/agent-base as a package instead of vendoring src/base/

swampratnz:
# ✅ The publish blocker is resolved — this installs from the public registry

`@swampratnz/agent-base@0.1.1` is **published to public npm** and this branch depends on it properly. The earlier "DO NOT MERGE until published" state is gone:

| was | now |
|---|---|
| `package.json` asked for `^0.0.1` — a version that never existed | `^0.1.1` |
| `package-lock.json` entry was **hand-written**, with a `resolved` URL and **no `integrity` hash** | written by a real `npm install`, with the registry tarball and its hash |
| `npm ci` failed on install, so every CI job was red before the first gate | `npm ci` succeeds from a clean tree (`rm -rf node_modules && npm ci` — verified) |
| `discord.js` pinned `^14.26.4` while the package requires `^14.27.0` → unsatisfiable tree | `^14.27.0`, resolved by npm, not hand-edited |
| a `postinstall` shim patched `node_modules` to add the package's missing subpath exports | 0.1.1 ships them; **the shim is deleted** |
| a cast worked around an uninhabitable `toolServerParts` field type | 0.1.1 made the module interface generic in the context type; **the cast is gone** |

The lockfile entry, which is the whole point:

```json
"node_modules/@swampratnz/agent-base": {
  "version": "0.1.1",
  "resolved": "https://registry.npmjs.org/@swampratnz/agent-base/-/agent-base-0.1.1.tgz",
  "integrity": "sha512-rVkNrmCdItkk+M3Mvb4w6djkT+EqZhwx+BTu2CZFHSaRWhxp8VFsHXHpddNALB11ezt3y4lEw0J/XPRs2DJ03Q==",
```

No auth or `.npmrc` is needed to install it — it is public on npmjs, which also settles the old "GitHub Packages vs npmjs" question.

---

# ⚠️ REQUIRED before deploy: two new environment variables

**The production `.env` must gain both of these, or the service refuses to start:**

```
DISPLAY_TIMEZONE=Pacific/Auckland
DISPLAY_LOCALE=en-NZ
```

This is not optional and not a default that quietly holds. The timezone and locale used to render every member-facing event time were hardcoded in `util/nzTime.ts`; they are the framework's config now, and the framework cannot know a deployment's timezone, so it defaults them to `UTC`/`en-GB`. An unset var re-renders every event time in UTC — invisible in tests and logs, and obvious only to a member who turns up an hour late. So `agentModule.init()` **asserts both values and refuses to boot** on anything else, and `tests/displayTime.test.ts` pins that the configured settings render byte-identically to the pre-flip `formatNzEventTime`.

`.env.example`, `docs/DEPLOYMENT.md` and the CI env blocks all set them. The production `.env` is the one copy this PR cannot reach.

---

## What this does

The dependency flip: community-agent stops carrying its own copy of the framework and consumes the package.

| | |
|---|---|
| `src/base/` | **deleted** — 121 files |
| `src/module/**` imports | repointed to `@swampratnz/agent-base/...` (~800 specifiers) |
| composition | `src/index.ts` → `createAgent({ modules: [nzCommunityModule] })` over ONE manifest, `src/module/agentModule.ts` |
| schema | this deployment contributes `src/module/storage/schema/` through `AgentModule.migrations`; base's fragments ship in the package |
| tests | 208 files repointed, 41 deleted (they live in agent-base), 3 added |

Commits, in order:

1. `Depend on @swampratnz/agent-base and delete the vendored src/base/`
2. `Repoint the test suite; drop the tests that now live in agent-base`
3. `Make the display timezone/locale and the two ingest URLs explicit env`
4. `Retarget the gates at a tree with no src/base/`
5. `Bring the docs up to a repo whose framework is a dependency`
6. `Install @swampratnz/agent-base from the registry, not a local tarball`
7. `Delete the postinstall shim that patched agent-base's exports map`
8. `Drop the toolServerParts cast and pin the manifest to this module's ToolContext`
9. `Resolve agent-base paths with the ESM resolver in the source-scanning tests`
10. `Correct the docs left describing a pre-flip tree and an unpublished package`

Commits 6–10 are the post-publish half; 1–5 are the flip itself, unchanged.

### Composition

Every content file stopped self-registering. The values are exported and named once, as data, in `src/module/agentModule.ts`: `notices`, `toolTiers`, `toolServerParts`, `flaggedToolPredicates`, `skills`, `promptSections`, `commands`, `defaultBadWords`, `personas`, `turnStateFinalizers`, `policyKeys`, `migrations`. `createAgent` performs them in its fixed order (plan → init → singletons → additive → readiness probe → migrate → start), so an incomplete composition is rejected with the process untouched instead of surfacing as a blank notice in front of a member.

The manifest is now declared with this deployment's own `ToolContext` as its type argument rather than left on the generic's `unknown` default. Pinning it typechecks every handler in the tool registry against the context `makeToolContext` actually builds, instead of admitting them on the default's documented bivariance — strictly more checking than the cast it replaces.

Two consequences worth knowing before you review the test diff:

- **No module-scope registry reads.** The pack is registered *after* every module is imported, so a module-scope `notice()` throws. Three derived constants in `agent/tools/helpers.ts` were removed for this reason; the function that used them already selected the notice at its call site.
- **Tests opt in per slice** via `tests/support/register*.ts` — one helper per manifest field, a 1:1 replacement for the side-effect imports tests already had, inserted in the same position and form (a suite that sets a dummy env before importing config still does).

### The two new notice ids

`gatedNoticeWithAdmins` (the dynamic admin-naming sentence base used to build inline) and `guidelinesHeading` (the literal `Community guidelines:`, which was duplicated across the router and all three adapters). Both are **base-text only, no `mi`/`plain` variants** — deliberately, because that is what the pre-flip behaviour was on every one of those paths. Adding variants would be a behaviour change, not this PR's business.

### The five dropped `TurnStateBag` keys

`communityTurnState.ts` keeps the `ToolServerTurnState` augmentation (the keys this deployment's tools WRITE) and drops the `TurnStateBag` half: agent-base declares those five itself now, because its router ships the handlers that READ them.

### `tests/support/agentBasePath.ts` (new)

Four tests read the framework's shipped source text off disk rather than importing it — "the Discord adapter never imports the Agent SDK", "the project repository has no outbound fetch call site", "`FEATURE_FLAG_MAP` covers every `*_ENABLED` var the config declares", "every schema `.sql` on disk is in a manifest". All four located the file with `createRequire(...).resolve()`.

That only ever worked because the deleted shim wrote the subpath entry as a **bare string** target, which matches under every condition including `require`. The published package maps its subpaths to a conditions object of `types` + `import` — correct for a `"type": "module"` package — and the CJS resolver then fails all of them with `ERR_PACKAGE_PATH_NOT_EXPORTED`. So they resolve through `import.meta.resolve` now, which asks for the condition the package actually publishes and is the same resolver every real import site in this repo already uses. That is the tests using the right resolver for an ESM-only dependency, not a workaround for a packaging defect.

Two of those failures were also masking others: the resolve threw mid-test, *before* the test's own cleanup, orphaning rows that then broke the next several row-count assertions in the same file.

---

## Env-surface delta

| Var | Before | After | Default preserved how |
|---|---|---|---|
| `DISPLAY_TIMEZONE` | hardcoded `Pacific/Auckland` in `util/nzTime.ts` | **new env var**; package default `UTC` | `.env.example`, deploy docs and the CI env blocks set `Pacific/Auckland`; `agentModule.init()` **refuses to boot** on anything else |
| `DISPLAY_LOCALE` | hardcoded `en-NZ` | **new env var**; package default `en-GB` | same, `en-NZ` |
| `DOCS_INGEST_INDEX_URL` | env, defaulted to `https://platform.claude.com/llms.txt` | env, **no default**; required when `DOCS_INGEST_ENABLED` | `.env.example` already set it explicitly to the same URL |
| `STATUS_CHECK_API_URL` | env, defaulted to `https://status.claude.com/api/v2/summary.json` | env, **no default**; required when `STATUS_CHECK_ENABLED` | same |

Only the display pair needs a hand on the production `.env` — see the banner above.

---

## Behaviour differences (all intended, none member-visible)

1. **Language/style hints are passed RAW.** The router used to pre-resolve the `mi`-or-nothing ternary; base passes the stored value through and lets the catalogue's *registered* axis values decide. `'auto'`/`'standard'` are not registered, so they render exactly what `undefined` did. Two `SECURITY:` tests in `router.test.ts` asserted the pre-resolution literally; they now assert the property that matters (never a registered variant for a caller who set none).
2. **Preference columns.** agent-base generalised `response_style_prefs.style` and `language_prefs.language` from value allowlists to shape checks. `src/module/storage/schema/80-preference-values.sql` restores this deployment's value lists under **new constraint names** (`*_allowed`), so a module never drops or reshapes a constraint base owns. On the production database the base-named inline CHECKs already exist and are untouched; the new ones are equivalent, so nothing observable changes. On a fresh database they are what narrows the shape check back down.
3. **`formatEventTime` now imports config.** `util/nzTime.ts` imported nothing; the package's `util/eventTime.ts` reads the display settings off the config singleton, so a *static* import of it validates the env. Three suites that set a dummy env in their body import it dynamically now.
4. **`discord.js` 14.26 → 14.27.** Required by agent-base's dependency range; nothing in `src/module/**` or `tests/**` touches an API that changed. The suite is the check and passes unchanged (2830/2830), including all 145 `discordAdapter.test.ts` cases.

---

## Lowered `security-floor` counts

`npm run test:security:fix --allow-lower`, 17 entries removed, **162 `SECURITY:` tests**. Every single one is a test file deleted in commit 2 because agent-base ships it **verbatim, with the identical count in its own floor** — this is coverage moving repositories, not coverage disappearing:

| file | count | file | count |
|---|---|---|---|
| `repository.test.ts` | 117 | `pendingAlertQueue.test.ts` | 3 |
| `whatsappCloudWire.test.ts` | 8 | `sanitizeName.test.ts` | 3 |
| `secrets.test.ts` | 5 | `whatsappLidMap.test.ts` | 2 |
| `knowledgeScope.test.ts` | 4 | `backgroundJobHealth.test.ts` | 1 |
| `moderationRepo.test.ts` | 4 | `healthState.test.ts` | 1 |
| `rateReservation.test.ts` | 4 | `knowledgeCandidateDedupDegradation.test.ts` | 1 |
| `memberId.test.ts` | 3 | `provenanceTrust.test.ts` | 1 |
| `pendingActions.test.ts` | 3 | `storageLifecycle.test.ts` | 1 |
| | | `whatsappWire.test.ts` | 1 |

`+1` for the new `tests/agentModule.test.ts`. Net 1260 → 1099, unchanged by commits 6–10.

**Nothing that tests THIS deployment lost a case.** Two files that look like duplicates were kept on purpose:

- `gatedNotice.test.ts` (7 SECURITY cases) — agent-base's own copy keeps only **1** of them, and the dynamic notice interpolates this module's pack template now, so the name-sanitisation, cap and wait-clause cases stay here.
- `schemaConstraintIdempotency.test.ts` — rewritten to run over the JOINED base+module concatenation `migrate()` actually applies, and gains a case asserting the module fragment never drops or re-adds a constraint base owns.

---

## Still wanted in agent-base

The two blocking defects are **fixed in 0.1.1** — subpath exports ship, and the module interface is generic in the tool-context type — so the shim and the cast are both deleted here. What remains:

### 1. `gatedNotice.test.ts` lost 6 of its 7 `SECURITY:` cases in the lift

agent-base's copy declares 1 `SECURITY:` test where this repo declared 7. The dropped cases cover `renderGatedNotice`'s name sanitisation, the empty-after-sanitise fallback, the `GATED_NOTICE_MAX_ADMIN_NAMES` cap and the wait clause's no-free-text pattern — all base behaviour, all currently unguarded upstream. This repo keeps its copy, so nothing is unguarded *here*, but agent-base should take them back.

### 2. Gap: no per-module config

`AgentModule` has no `configSchema` field (plan §3 has one; `createAgent` does not implement it), so a new env var for a community feature is an agent-base change plus a version bump. `docs/agents/recipes.md`'s config recipe is marked ⚠️ UPSTREAM accordingly. This is the last item on Phase 3's owed list.

### 3. Nice-to-have for 0.1.2: no `require`/`default` condition on the subpath exports

The wildcard subpath entry resolves to a `{ types, import }` conditions object only, so the CJS resolver — `createRequire(...).resolve('@swampratnz/agent-base/config.js')`, and anything else asking for the `require` condition — gets `ERR_PACKAGE_PATH_NOT_EXPORTED` rather than a path. This is *defensible* for a `"type": "module"` package and is **not blocking**: the consumer-side fix is to use the ESM resolver, which is what `tests/support/agentBasePath.ts` now does, and no runtime import site was affected. But Node ≥22 can `require()` ESM, so adding a `default` condition alongside `import` would make the package resolvable from either loader at no cost. Filed here so it is a decision rather than a surprise for the next consumer.

---

## Gate numbers

Postgres 16 + pgvector, a dedicated database (`ca_960`) migrated from scratch, one suite at a time, against the **published** package installed via `npm ci`:

| gate | result |
|---|---|
| `rm -rf node_modules && npm ci` | ✅ 356 packages, 0 errors — the command that was failing |
| `npm run migrate` | ✅ base + module fragments, one query, `DATABASE_URL` only |
| `npm run typecheck` (+ `typecheck:tests`) | ✅ clean, with no `toolServerParts` cast |
| `npm run lint` | ✅ clean |
| `npm run format:check` | ✅ clean |
| `npm test` | ✅ **2830 pass, 0 fail** |
| `npm run test:security` | ✅ **1099 `SECURITY:` tests ran, manifest total 1099 across 149 files** |
| `npm run build` | ✅ `dist/module/storage/schema/` matches the manifest (1 fragment) |
| `npm run context:check` | ✅ 24 required paths, 65 entries, sorted, no stubs |
| `npm run imports:check` | ✅ 66 files obey the composition-direction rules |

`imports:check` would have become a permanently-green no-op (its subject was an import edge out of a deleted directory), so it is **retargeted, not left as decoration** — see commit 4 and the "I got an `imports:check` failure" recipe.

---

## Docs corrected in commit 10

The automated review flagged one stale line and there were six more of the same kind. `docs/agents/module-map.md`'s `src/module/` entry still read *"Free to import base; nothing in base may import back"* — a one-way rule between two directories, one of which no longer exists. It now states what `imports:check` actually enforces. Also fixed: the gate description naming `src/base/` as a scanned root, the "not complete" note pointing at `src/base/agent/`, two entries pointing at `src/base/jobs/`, CLAUDE.md's migrate boot slice, and CLAUDE.md's secrets note contrasting `whatsapp-auth/` with a `src/base/auth/` that is gone (the real invariant there is that `.gitignore`'s patterns are anchored to the repo root).

The remaining `src/base/` mentions are deliberate — the rule that it must never come back, and the gate message that fires if it does.

---

## Before this can merge

1. ~~Publish `@swampratnz/agent-base`~~ — done, 0.1.1 on public npmjs.
2. ~~Real lockfile entry with an `integrity` hash~~ — done, commit 6.
3. ~~Drop the postinstall shim and the `toolServerParts` cast~~ — done, commits 7 and 8.
4. **Confirm CI is green.** Every previous run died at `npm ci`; this is the first branch state that can actually reach the gates.
5. **Add `DISPLAY_TIMEZONE=Pacific/Auckland` and `DISPLAY_LOCALE=en-NZ` to the production `.env`** — see the banner above. Without them the service refuses to start.
6. Flip agent-base's `AGENT_BASE_CANARY_ENABLED` repository variable on, so the canary stops measuring the gap and starts guarding it.

`Closes #—` deliberately omitted: this is the Phase 3 consumer half, tracked by `docs/AGENT-BASE-PLAN.md` §Phase 3, which this PR updates with exactly what remains.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hkgg61b9V5K1tDwE6t6kUs)_


swampratnz:
CI is red on all three jobs at `npm ci`, and it is expected at this stage rather than a defect in the diff. Diagnosis, so the rerun attempts stop:

```
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json are in sync.
npm error Invalid: lock file's discord.js@14.26.4 does not satisfy discord.js@14.27.0
```

Three causes, all upstream of this branch:

1. **The lockfile names `@swampratnz/agent-base@0.0.1`** — a version that has never existed on the registry (`0.1.0` is what shipped), hand-written without an integrity hash because the package was unpublished when this branch was built. Deliberately not fabricated, which was the right call; it just cannot install.
2. **`@swampratnz/agent-base@0.1.0` requires `discord.js@^14.27.0`** while this repo pins `^14.26.4`. That is the error above — an unsatisfiable tree, not drift in this diff.
3. **The package exposes no subpath exports** (`exports` covers only `.` and `./package.json`), which is why this branch carries `scripts/patch-agent-base-exports.mjs`, a postinstall that patches `node_modules`. That is not something to merge: it mutates an installed dependency on every install.

Being fixed properly upstream in `agent-base` rather than worked around here — subpath exports, the uninhabitable `ToolServerParts<never>` slot, and a **security-coverage regression** the flip pass found and reported honestly: `tests/gatedNotice.test.ts` arrived in agent-base having lost 6 of its 7 `SECURITY:` cases (admin-name sanitisation, the name cap, the wait-clause pattern). The floor gate could not catch that, because the file was new in that repo and its count was whatever it arrived with. Every other lifted test file is being audited the same way against `8e5d0db`.

When `0.1.1` publishes, this branch gets: a real `npm install @swampratnz/agent-base@0.1.1` with a genuine integrity hash, `discord.js` moved to `^14.27.0`, and the postinstall shim deleted.

Holding this PR until then — reruns will keep failing identically until the dependency exists.

---

swampratnz:
Agreed on the verdict — this is blocked on an out-of-repo publish, not on anything reviewable here. Two corrections and a status update, so the "registry decision" framing doesn't send anyone chasing a question that is already settled.

**There is no npmjs-vs-GitHub-Packages decision outstanding.** Item 5 in the PR body is wrong, and I verified it rather than repeating it: `agent-base`'s `publishConfig` reads

```json
{"access": "public", "registry": "https://registry.npmjs.org"}
```

The GitHub Packages reading came from a tree predating agent-base's trusted-publishing change. The registry decision was made and recorded in `docs/RELEASING.md` there: **public npmjs**, chosen because GitHub Packages' npm registry requires an auth token to *install* even for public packages, which would put a credential expiry between the production host and `npm ci`.

**The publish has already happened.** `@swampratnz/agent-base@0.1.0` is live, MIT, and installs from a clean directory with no `.npmrc` and no token — verified by downloading the tarball (420 files, all 26 `.sql` fragments present) and installing it unauthenticated. Publishing now runs by OIDC trusted publishing, so no npm secret exists in either repo.

What actually blocks this PR is narrower than "a registry decision": `0.1.0` ships no subpath exports, requires `discord.js@^14.27.0` against this repo's `^14.26.4` pin, and this branch's lockfile names a placeholder `0.0.1`. `0.1.1` is in flight to fix the exports and the uninhabitable `ToolServerParts<never>` slot; this branch then gets a real install with a genuine integrity hash, the `discord.js` bump, and `scripts/patch-agent-base-exports.mjs` deleted.

**On `gatedNotice.test.ts`** — thanks for confirming the floor delta is a relocation. Worth stating the general hazard plainly, because it is a gate blind spot rather than a one-off: the security floor protects against deleting cases *within* a repo, but a file arriving in a *new* repo gets whatever count it shows up with, so cases can be lost in transit and the manifest will record the smaller number as correct. That is exactly what happened to this file. `0.1.1` restores the six missing cases upstream, and every other lifted test file is being compared case-for-case against `8e5d0db`.

Taking the stale `src/module/` wording in `docs/agents/module-map.md` in the same push as the dependency bump — good catch.

---

swampratnz:
Applied `allow-security-floor-lower`. This is the gate's designed escape hatch, so here is the justification it requires, verified rather than asserted.

**The gate's complaint:**
```
total SECURITY: test count DROPPED 1260 → 1099 vs the PR base — the NET number
of SECURITY: tests fell (a deletion, or a rename that dropped one).
```

**What the delta actually is**, computed by diffing `tests/security-floor.json` against `origin/main`:

| | files | cases |
|---|---|---|
| removed | 17 | 162 |
| added | 1 | 1 (`agentModule.test.ts`) |
| **changed in place** | **0** | **0** |

That last row is the one that matters. Erosion — a file quietly losing a case or two — shows up as a changed-in-place count. There are none: every delta is a whole file leaving, which is what relocation looks like and what partial loss does not.

**Where those 17 files went.** Each is now in `swampratnz/agent-base`, and I compared its count there against its count here before the removal:

```
relocated with identical count: 17/17
```

No mismatches, none absent. The list, with cases: `repository` 117, `whatsappCloudWire` 8, `secrets` 5, `knowledgeScope` 4, `moderationRepo` 4, `rateReservation` 4, `memberId` 3, `pendingActions` 3, `pendingAlertQueue` 3, `sanitizeName` 3, `whatsappLidMap` 2, and six single-case files (`backgroundJobHealth`, `healthState`, `knowledgeCandidateDedupDegradation`, `provenanceTrust`, `storageLifecycle`, `whatsappWire`).

**Independent confirmations**, since one check is not evidence:
1. The automated reviewer on this PR verified the drop as relocation against the deleted files in commit 2.
2. agent-base's `0.1.1` pass diffed **test names**, not just counts, across all 44 lifted files against `8e5d0db` — and found exactly one file that had lost anything (`gatedNotice.test.ts`, 7 → 1). That one was **deliberately kept here** rather than removed, and its missing six cases were restored upstream in `0.1.1`. It is the reason this PR's removal list has 17 entries and not 18.

**A gate limitation worth recording**, because it bit us once already: this floor protects against deleting cases *within* a repo. A file arriving in a *new* repo gets whatever count it shows up with, so cases can be lost in transit and the receiving manifest records the smaller number as correct. Only a name-level diff against the source catches that. That is how `gatedNotice.test.ts` lost six cases silently, and why the audit above was done by name rather than by count.

---

swampratnz:
Merging. Green on `c2e1366`: `build` ✅, `lint` ✅, `security-invariants` ✅. The red `review` check is the turn-cap artifact — that same reviewer returned a full **LGTM** on the previous head, so its substance is on the record.

**⚠️ Before the next deploy**, production `.env` needs two new lines:

```
DISPLAY_TIMEZONE=Pacific/Auckland
DISPLAY_LOCALE=en-NZ
```

These are required, not optional. The framework defaults to UTC and `en-GB`, so without them every event time and date the bot renders would silently shift. `agentModule.init()` refuses to boot rather than serve wrong times — the deploy stops instead of misbehaving, which is the failure mode you want, but it does mean the deploy stops.

**Three defects were found and fixed on the way here, none in the product code** — all in the gates, all invisible until an unusual condition hit them:

1. **The floor override leaked into the tests asserting it.** `ALLOW_SECURITY_FLOOR_LOWER` is a job-level env var, and `securityGateWrite.test.ts` spawned the gate with `{ ...process.env, ...env }` — so on any labelled PR the "must block a lowering" cases ran against a gate with that guard switched off. Latent since #851, and only a labelled PR could ever trigger it.
2. **The gate discarded the output naming the failure.** It wrote the child's TAP with `process.stdout.write` then called `process.exit()`; on a pipe that write is still buffered, so ~964 of 1,099 lines went in the bin — every `not ok`, the summary, and its own error message. CI showed a bare "exit code 1" ending mid-token after a *passing* test. Now `writeSync`, which leaves control flow alone.
3. **Two gate tests raced over the real repo tree.** The no-`src/base/` case created one under the actual root; `contextPack.test.ts` runs the context checker against that same tree with no `--root`, and an undescribed `src/base/` fails it. Different files, so `node:test` runs them in parallel — the victim would have been an unrelated PR. Now driven against a temp fixture via a new `--forbid-base` flag, verified in all three modes.

Thanks to the automated reviewer for catching (3) — it was flagged as theoretical, and turned out to have a concrete counterpart test waiting to lose the race.

---

PR #1232, opened 2026-08-30
Build the exactly-at-threshold alert fixture before the job reads its clock

### LINKED ISSUE #1231: Flaky: the exactly-at-threshold stale-alert test has zero slack against the job's clock
## What

`tests/accessRequestStaleAlert.test.ts` — *"an access request exactly at the threshold DOES trigger the alert"* — dates its fixture **inside** the lazy `listPendingAccessRequests` callback:

```ts
const listPendingAccessRequests = async () => [
  accessRequest({ ageHours: ACCESS_REQUEST_STALE_ALERT_THRESHOLD_HOURS }),
];
```

The job (`src/module/accessRequestStaleAlert.ts:98`) captures its clock **first** and only then awaits that callback:

```ts
const now = Date.now();
const requests = await listPendingAccessRequests();
```

So the fixture is stamped at `now + δ` for δ ≥ 0, and the measured age is `THRESHOLD − δ`. The comparison is `>=`, so any δ ≥ 1 ms makes the request **not stale** — no DM is sent and the test's `assert.equal(dms.length, 1)` fails.

## Why this one is sharper than the two already fixed

This is the same root cause as #1071 (`appealStaleAlert`) and the `rosterStaleAlert` multi-platform case fixed in #1229, but those fixtures carry 30h+ of margin, so the race only ever **misrendered the hour count** (`100h` → `99h`). This fixture sits *exactly* on the `>=` boundary and has no slack at all, so the race **inverts the very behaviour the test asserts**.

Padding the age is not a fix — the boundary is the point of the test. The fix is to build the fixture eagerly, which makes the age `THRESHOLD + δ`.

## Evidence

Against the real test, with an artificial 2 ms gap in the fixture clock standing in for a loaded runner:

| | result |
|---|---|
| lazy fixture (today) | `not ok 5 — exactly-at-threshold counts as stale (>= comparison)`, 14 pass / 1 fail |
| eager fixture | 15 pass / 0 fail |

It passes on an idle machine because both `Date.now()` calls land in the same millisecond, which is why this has survived unnoticed.

## Scope

A sweep of every `const now = Date.now()`-then-`await` job in `src/module/` finds six, all in the stale-alert family. Their test files were each checked for boundary-sensitive fixtures:

- `accessRequestStaleAlert` — the case above, plus a `THRESHOLD - 1` case (safe).
- `appeal` / `knowledgeCandidate` / `report` / `roster` / `suggestion` — `THRESHOLD - 1` only, which is safe in the failing direction: a later stamp makes the row *younger*, so it stays under the threshold and the "no alert" assertion still holds.

Every exact-hour-string assertion in those files already builds eagerly (each carries a comment naming the #1071 precedent).

Separately worth noting, not fixed here: five of the six siblings have **no** exactly-at-threshold test at all, so the `>=` boundary is pinned in only one of six jobs that share an identical threshold/latch shape.

swampratnz:
Closes #1231

## Summary

`tests/accessRequestStaleAlert.test.ts` — *"an access request exactly at the threshold DOES trigger the alert"* — dated its fixture **inside** the lazy `listPendingAccessRequests` callback. The job (`src/module/accessRequestStaleAlert.ts:98`) captures `const now = Date.now()` **first** and only then awaits that callback, so the row was stamped at `now + δ` and its measured age was `THRESHOLD − δ`. The comparison is `>=`, so any millisecond of clock advance made the request not stale: no DM, and the test's `assert.equal(dms.length, 1)` failed.

Same root cause as the `appealStaleAlert` fix in #1071 and the `rosterStaleAlert` multi-platform fix in #1229, but **sharper**. Those fixtures carry 30h+ of margin, so the race only ever misrendered the hour count (`100h` → `99h`). This one sits exactly on the `>=` boundary with no slack at all, so it inverts the behaviour the test asserts.

Padding the age was not available as a fix — the boundary *is* the point of the test. Building the fixture eagerly makes the age `THRESHOLD + δ`, which satisfies `>=` on every run.

### Sweep

Every `const now = Date.now()`-then-`await` job in `src/module/` was enumerated: six, all in the stale-alert family (`accessRequest`, `appeal`, `knowledgeCandidate`, `report`, `roster`, `suggestion`). This is the **only** exactly-at-threshold fixture among them. The other five test `THRESHOLD - 1` only, which is safe in the failing direction — a later stamp makes the row *younger*, so it stays under the threshold and the "no alert" assertion still holds. Every exact-hour-string assertion in those files already builds eagerly.

Worth flagging, deliberately **not** done here: five of those six siblings have no exactly-at-threshold test at all, so the `>=` boundary is pinned in only one of six jobs sharing an identical threshold/latch shape. That is a coverage gap, not a flake, and adding five tests is outside a flake repair.

## Security / privacy impact

**None.** Test-only, one file. No change to `src/`, tool gating, RBAC, outbound filtering, data-access scope, schema, config, or stored data. No `SECURITY:` test added or removed, so `tests/security-floor.json` is untouched.

This does *increase* assurance on one security-adjacent behaviour: the corrected test now reliably pins that a request sitting exactly on the staleness threshold **does** alert, rather than passing or failing on clock luck.

## How verified

Full gate on the merged tree against a **freshly created** Postgres 16 + pgvector database:

- `npm test` — **3741 pass, 0 fail, 0 skipped**
- `npm run test:security` — **1506 pass, 0 fail**, manifest exact-match
- `npm run typecheck` (incl. `typecheck:tests`), `lint`, `format:check`, `context:check`, `imports:check`, `migrate`, `build` — all clean

The fix is demonstrated rather than asserted. Injecting a 2 ms gap into the fixture clock, standing in for a loaded runner:

| fixture form | result |
|---|---|
| lazy (before) | `not ok 5 — exactly-at-threshold counts as stale (>= comparison)` — 14 pass / 1 fail |
| eager (after) | 15 pass / 0 fail |

It passes on an idle machine because both `Date.now()` calls land in the same millisecond, which is why it has survived unnoticed.

No CHANGELOG entry: test-only, so this is internal under the changelog-coverage skip-ledger convention.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QgYZcxN4CTBbCeLVTGu8aH)_

### DIFF
--- a/tests/accessRequestStaleAlert.test.ts
+++ b/tests/accessRequestStaleAlert.test.ts
@@ -210,9 +210,20 @@ test('makeDefaultAccessRequestStaleAlertRun: a pending-access-request set with n
 
 test('makeDefaultAccessRequestStaleAlertRun: an access request exactly at the threshold DOES trigger the alert', async () => {
   const { adapter, dms } = makeAdapter();
-  const listPendingAccessRequests = async () => [
-    accessRequest({ ageHours: ACCESS_REQUEST_STALE_ALERT_THRESHOLD_HOURS }),
-  ];
+  // Built EAGERLY, before runOnce() captures its clock — the same ordering fix
+  // as the stale set above, but load-bearing for a sharper reason: this
+  // fixture sits EXACTLY on the `>=` boundary, so it has no slack at all. The
+  // job reads `Date.now()` first and only then awaits
+  // listPendingAccessRequests(), so a fixture dated inside that lazy callback
+  // is stamped a hair LATER than `now`, making the measured age
+  // `THRESHOLD - delta` — just under the threshold, so the request is not
+  // stale and no DM is sent. The other alert tests carry 30h+ of margin and
+  // only misrender the hour count; this one inverts the very behaviour it
+  // asserts. Building first makes the age `THRESHOLD + delta`, which is
+  // >= threshold on every run. Padding the age instead would destroy the
+  // point of the test, which is the boundary itself.
+  const atThreshold = [accessRequest({ ageHours: ACCESS_REQUEST_STALE_ALERT_THRESHOLD_HOURS })];
+  const listPendingAccessRequests = async () => atThreshold;
   const listAdminIdentities = async () => admins([{}]);
   const runOnce = makeDefaultAccessRequestStaleAlertRun(
     [adapter],


swampratnz:
**`lint` is red for an infrastructure reason, not this diff.**

It died inside `npm ci`, in `onnxruntime-node`'s postinstall script, fetching a prebuilt binary:

```
npm error path .../node_modules/onnxruntime-node
npm error command sh -c node ./script/install
npm error AggregateError [ETIMEDOUT]:
npm error   Error: connect ETIMEDOUT 150.171.109.119:443
npm error   Error: connect ENETUNREACH 2603:1061:14:73::1:443
```

That is dependency installation, before any lint body ran. This PR changes one test file (`tests/accessRequestStaleAlert.test.ts`) and touches no manifest, lockfile, or config, so it cannot affect what `npm ci` does.

The decisive evidence is within this same run, on the same commit and lockfile: `review` installed and passed, and `build` + `security-invariants` both got past `npm ci` and are still executing. Three installs from one commit, one timed out. That is a transient runner network fault, not a reproducible break.

**No manual re-run from me, deliberately.** `ci-retry.yml` triggers on CI-run completion at `run_attempt < 2` with `failed-jobs-only: true`, so lint gets its free machine retry automatically once `build` and `security-invariants` finish and the run concludes. Re-running by hand now would race that and spend the same one-shot budget twice — and per that workflow's own contract, `pipeline-pr-autofix.yml` only engages from attempt 2, precisely so a flake costs zero agent tokens.

If the retry fails a second time, that is a real signal and I'll root-cause it rather than re-run again.

For the record, the rest of the PR is already settled: `review` posted an LGTM on this head, and the full gate was green locally against a freshly created Postgres 16 + pgvector database — `npm test` 3741/3741, `npm run test:security` 1506/1506, plus typecheck, lint, format:check, context:check, imports:check, migrate and build all clean.


---

Issue #1328, opened 2026-09-08
react_to_message is the actual last member-tier tool with zero te reo Māori support — two prior "last one" fixes (#1147, #1176) both missed reactions.ts


swampratnz:
## Problem

`react_to_message` (`src/module/agent/tools/reactions.ts`) is a `minTier: 'member'` tool, and every one of its six `text(...)` reply strings is hardcoded English with no `getLanguagePreference` lookup: the success ack (`Reacted ✅.`), the daily-limit refusal, the "not available on this platform" refusal, the "no message to react to" refusal, the "never seen in this conversation" refusal, and the failure message.

This matters because the standing `'mi'` preference (`set_language_preference`, #189) is meant to be a blanket guarantee — every future reply, in every conversation — not a per-tool opt-in. Two prior proposals each explicitly claimed to close out the *entire* member-tool surface:

- #1147, "feedback.ts and reportsMember.ts are the **last two** member-tool files with zero te reo Māori support"
- #1176, "memory.ts is **the last** member-tier tool file with zero language-preference handling"

Both shipped, and both were wrong about completeness — `reactions.ts` was never touched by either sweep. I verified this directly by reading `src/module/agent/tools/index.ts`'s tier map and every file it lists under `minTier: 'member'` (11 files total: `selfService.ts`, `social.ts`, `knowledgeMember.ts`, `prefs.ts`, `projectNotes.ts`, `reactions.ts`, `reportsMember.ts`, `info.ts`, `memory.ts`, `digestMember.ts`, `feedback.ts`) against which ones call `getLanguagePreference`. Exactly two don't: `prefs.ts` and `reactions.ts`.

`prefs.ts` is a deliberate exception, not a gap: its two tools (`set_response_style`, `set_language_preference`) reply to the member in the *instant* they're changing that very setting, so confirming in a language they may not have chosen yet is a defensible, different design question — not the same "reply ignores an already-standing preference" bug. `reactions.ts` has no such excuse: a member with a long-standing `'mi'` preference gets English every single time they react to a message.

Reach: `react_to_message` is available to every member on both Discord and WhatsApp (Baileys + Cloud) per its `requiresCapability` union — one of the most frequently-invoked member tools in the surface, precisely because it's the low-friction "got it/noted" acknowledgement path the tool description advertises.

## Proposed approach

[...]

Issue #53, opened 2026-07-03
Proposal: give the research loop read access to an anonymised community-context export so proposals are grounded in what the community actually needs


swampratnz:
## Problem

The research loop files proposals to improve the bot, but today it works from the repo, issues, and VISION.md — it is **blind to the actual community**. It can't see that members keep asking the same three questions, or that a topic dominates the channels. The richest signal for "what should the bot do next" lives in the interaction corpus and (once built) the context digests (#context-builder proposal), but the research loop has no path to it. Closing that loop is what makes the system genuinely self-improving: distilled community need → grounded proposal → build → deploy.

## Proposed approach

One PR — a **one-way, anonymised export** the research loop reads, with a hard privacy boundary between the DB and anything that reaches GitHub:

1. **Exporter** (`src/context/export.ts`): renders the `context_digests` (topics, recurring-question counts, summaries) into a single artifact — `docs/COMMUNITY-CONTEXT.md` (committed) or a build-time JSON the research prompt ingests. Regenerated by the context-builder run.
2. **Anonymisation is the whole point** and must be enforced in code, not by prompt:
   - **No raw message content, no user identifiers, no display names, no conversation ids.** Only aggregate topic labels, counts, and model-written summaries.
   - A **k-anonymity floor**: a topic/question cluster is only exported if it's backed by ≥K distinct users (e.g. K=3), so a single person's messages can never become an identifiable line in a world-adjacent file. Clusters under the floor are dropped (and the drop is logged, not silently hidden).
   - A scrub/allowlist pass on the generated text (strip anything resembling emails, phone numbers, @handles, URLs with user tokens) before it's written — reuse/extend the outbound `redactSecrets` machinery.
3. **Research-loop wiring**: the research Routine's prompt gains a step to read the export and cite it as evidence in proposals (fits VISION.md's evidence rubric). The loop stays **issues-only and read-only** per pipeline ownership — it consumes the export, never the DB directly (it has no DB access, and this proposal must not give it any).
4. **Freshness**: the export carries a generated-at timestamp and the window it covers, so a stale export is visible.

## Alternatives considered

- **Give the research loop direct DB / recall access** — rejected outright. It would put raw member content (PII) into a loop that writes world-readable GitHub issues, and widen the loop's surface far beyond issues-only. The anonymised, k-floored, scrubbed export is the safe interface.
- **Export raw digests without the k-floor** — a niche topic raised by one identifiable person could leak into an issue; the floor is non-negotiable.
[...]

PR #1394, opened 2026-09-11
feat: member web research + link summaries (off by default; needs agent-base 0.6.5)

swampratnz:
> **Depends on `@swampratnz/agent-base` 0.6.5, now published** (swampratnz/agent-base#55, merge `981bd5d`; npm `gitHead` verified to match). This branch installs it (package.json + lock). The earlier red CI (`build`, `security-invariants`) was only this branch running against 0.6.4 before the release.

## Why

A 30-day read of the production database turned up a gap. Of 56 member questions, 49 went unanswered: they were about current events (releases, pricing, outages) or about a link someone had just posted, and the curated knowledge base can't answer either kind. Its median best-match cosine distance on those questions was 0.654. Members have no web access by design (WebSearch is admin+, WebFetch is banned for every tier), so "I don't know" was the only honest answer Dave had.

## What

Two member tools, **both shipped OFF** (`WEB_RESEARCH_ENABLED`, `LINK_SUMMARY_ENABLED`, literal `"true"` only):

### `web_research`: look it up, with sources, when the knowledge base has nothing
- **Mechanically gated on a knowledge miss, keyed to the question.** `knowledge_search` sets `turnState.knowledgeSearchMissed` when it returns no relevant or lexical hit, and `web_research` refuses unless that flag is set. It *also* runs its own pre-check (`knowledgeCovers`, the same semantic-floor-then-lexical rule) on the exact question it was given, before dedup and quota, and refuses if curated knowledge covers it. If that lookup errors, it fails closed. That second check was added after review: the turn flag alone is sticky, so an unrelated, deliberately-missing search could otherwise unlock research on a question the knowledge base answers. It's pinned by 3 SECURITY tests (semantic-covered, lexical-covered, lookup failure).
- **Isolated sub-turn.** One `query()` whose prompt is *only the question*: no history, no names, no conversation id, no module MCP servers, `settingSources: []`, tools exactly `['WebSearch']`, `WebFetch`/`Task` disallowed, `maxTurns` from config, 90 s abort. A prompt injection planted in a search result lands in a context with nothing to exfiltrate and no tool that could send anything anywhere. The canary test proves the conversation id and display name never reach that prompt.
- **Structured output.** JSON schema `{found, answer, sources[]}`, parsed strictly. Only `https` sources are kept (max 6, titles flattened). The answer comes back inside `untrusted()` quarantine, and `found: false` becomes a plain "no credible answer, say so".
- **Metered and watched.** Per-caller daily cap (default 5) plus a 10-minute dedup. Cost goes to `background_job_costs` as `web_research`, which agent-base's cost-spike alert now tracks.

### `summarize_link`: "TLDR?" on a link a person posted
- **Selection, never composition.** The URL must appear verbatim in an **inbound** message in the caller's **own** conversation within the lookback window (default 24 h). Platform and conversation id come from the envelope, never from the model. The tool fetches the *posted* normalized href, with that host as the entire `safeFetch` allowlist, so appending a query string, a path or a fragment is refused. A URL only the bot said never counts, which closes the say-then-fetch exfiltration loop.
- **All `safeFetch` guarantees apply:** SSRF denylist, DNS pinning, a per-hop host re-check (a redirect off the posted host is refused with honest advice), and byte/time/redirect ceilings shared with `FETCH_PAGE_*`. https only.
- **Metered.** Per-caller daily cap (default 20) plus a 5-minute dedup. Logs record the host and outcome kind only, never the URL path or the page.

### Also
- Both tools honour a standing te reo Māori preference (the #1328 coverage rule). Their output is English model-facing content, not fixed strings, so a `mi` caller gets a relay-language note rather than a machine translation.
- `community_info` mentions both capabilities, **folded into the existing knowledge bullet** (consolidate, don't append). The three reply-length caps got small documented bumps for that clause (member 2510→2600, admin 5170→5230, super_admin 5820→5880).
- `.env.example`, `docs/ARCHITECTURE.md` (tool matrix), `docs/SECURITY.md` §1 (both surfaces, with residual risks), `CHANGELOG.md`, `docs/agents/module-map.md` (🔒 entries).

## Consuming agent-base 0.6.5: fixture updates this release requires

0.6.5 also ships agent-base's unreleased audit remediations (#53/#54). One of them makes `slashDispatch` check the caller's block status and resolve their role **before** any handler runs. That broke 4 `discordSlashCommands` tests. I measured it on a clean `main` with the 0.6.5 pack, and they pass on 0.6.4:
- `mockPool` answered every `FROM blocked_users` query with `/blockedlist`'s fixture rows, so the gate read the seeded caller as blocked. It now answers the gate's `SELECT 1 FROM blocked_users` point lookup separately.
- The two `/blockedlist` SECURITY assertions now target the list read they actually protect (`listBlockedUsers`), not the gate's lookup of the caller's own row.
- The `/kb` ordering pin now allows exactly one read before role resolution: the base block-list gate.

Plus the expected pins: `feature_flags` anti-drift count 49→51, registry count 132→134, and the new tools added to the flags-off exclusion lists in the six `buildQueryOptions` baseline suites.

## Residual risks (also in SECURITY.md)
- **web_research:** a poisoned search result can still produce a wrong *answer*. It reaches the member quarantined and with sources, and the tool can't act on it.
- **summarize_link:** a member can make the bot fetch any public https page by posting it first. That's the intended feature, bounded by the posted-host allowlist, the SSRF floor and the per-caller cap. The page content is untrusted and returned quarantined.

## Verification (Windows checkout, Linux CI is authoritative)
- `typecheck` (src + tests ratchet) ✓, `build` ✓, `context:check` ✓, `imports:check` ✓, ESLint ✓, Prettier ✓ on every changed file.
- New suites: `webResearchTool` (11 SECURITY) and `linkSummaryTool` (11 SECURITY), all passing; `security-floor.json` updated.
- Full suite: 4188 tests, 3275 pass, 908 skipped (DB-gated), **5 fail**, all in `importDirection` (below).
- Known local-only red: `importDirection.test.ts` (5 tests) fails identically on clean `main`, because its expected output uses `/` paths and Windows prints `\`.

## Needs a human
- **Te reo Māori:** the `community_info` knowledge-bullet line is best-effort and needs a fluent speaker's review before this ships.
- **Enabling:** both flags stay off until the owner says otherwise. Suggested rollout: enable `LINK_SUMMARY_ENABLED` first (no model spend), watch `web_research` cost on the daily digest after enabling it.

https://claude.ai/code/session_01ULGfkzvZLCXrDRiz9CsMH7


swampratnz:
Addressed the review in the latest commit.

**Knowledge-first gate (design gap): fixed.** The sticky `turnState.knowledgeSearchMissed` flag is kept as the "knowledge_search was consulted this turn" precondition. On top of it, `web_research` now runs `knowledgeCovers()` on the *exact question it was given*. It uses the same rule as `knowledge_search` (a semantic hit at or above the relevance floor, otherwise a lexical hit on below-floor candidates) and runs before dedup and quota, so a refusal costs the member nothing. If that lookup errors, it fails closed with a do-not-guess instruction. An unrelated, deliberately-missing search can no longer unlock research on a question the curated base answers. Three new `SECURITY:` tests cover semantic-covered, lexical-covered and lookup-failure; the floor moves 8 → 11, and SECURITY.md and ARCHITECTURE.md are updated.

**`discordSlashCommands` fixtures ("speculative against an unreleased API"): not speculative.** They were written against a local `npm pack` of agent-base#55's exact head, the same artifact the canary builds. That PR's own canary run shows exactly these 4 tests plus the feature-flag pin as the only failures of community-agent `main` against it, and the query shape mocked here (`SELECT 1 FROM blocked_users WHERE platform = $1 AND external_id = $2`) is copied from `isUserBlocked` in that source.

**CI red:** still only the unpublished 0.6.5. `build` can't typecheck `config.webResearch`/`config.linkSummary` against the installed 0.6.4, and the three `security-invariants` files fail for the same reason (they pass 19/19 locally on 0.6.5). This stays draft until 0.6.5 is tagged.

swampratnz:
Follow-up on the latest automated review. It ran against a stale description; the body is now refreshed.

- **agent-base 0.6.5 is published.** It was tagged from merge `981bd5d`, and npm's `gitHead` matches that commit. This branch pins it exactly (`dependencyPins` requires a bare version), and `security-invariants` is green on it.
- **`discordSlashCommands` fixtures are now verified against the real package.** They run in this PR's CI against the published 0.6.5, not only against the local pack.
- **Link-history window, newest or oldest 500:** it's the newest. `recentConversationHistory` runs `ORDER BY created_at DESC LIMIT $4` and then reverses. Overflow in a busy window drops the *oldest* links, never a just-posted one. I added a comment on `PROVENANCE_SCAN_LIMIT` saying so.
- **Citation comment nit: fixed.** The https-only filter and the 6-source cap on `SOURCES` are the actual control. The comment no longer claims the prompt's citation rule (scoped to `knowledge_search`'s source clause) covers it.
- **Te reo Māori line:** still needs a fluent speaker's review, as flagged.

